PR #43, opened 2025-10-27
feat: mark interceptions as completed

pawbana:
Adds logic that marks interceptions as completed.

Needed for https://github.com/coder/internal/issues/1051

### DIFF
--- a/bridge_integration_test.go
+++ b/bridge_integration_test.go
@@ -1155,16 +1155,15 @@ func (m *mockRecorderClient) RecordToolUsage(ctx context.Context, req *aibridge.
 	return nil
 }
 
-// verify all interceptions has been marked as completed
+// verify all recorded interceptions has been marked as completed
 func (m *mockRecorderClient) verifyAllInterceptionsEnded(t *testing.T) {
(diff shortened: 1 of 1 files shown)


bridge_integration_test.go
```diff
@@ -1131,6 +1155,19 @@ func (m *mockRecorderClient) RecordToolUsage(ctx context.Context, req *aibridge.
...
+// verify all interceptions has been marked as completed
+func (m *mockRecorderClient) verifyAllInterceptionsEnded(t *testing.T) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	if len(m.interceptions) == 0 {
+		t.Errorf("HMMM")
+	}
+	require.Equal(t, len(m.interceptions), len(m.interceptionsEnd))
+	for _, intc := range m.interceptions {
+		require.Contains(t, m.interceptionsEnd, intc.ID)
```
deansheather: Probably should add a message to this require call so it makes more sense what it's even doing. Perhaps also a `t.Helper()` call at the top of this method

Issue #25, opened 2025-10-14
Expose Prometheus metrics


ibetitsmike:
Required:
- all partitioned by provider, model, client, user_id, endpoint, streaming (where applicable)
- all the metrics collected in the [telemetry packet](https://github.com/coder/coder-telemetry-server/issues/26)
- interception_duration
- passthrough count -> partitioned by provider, endpoint, user_id

We'll import the prometheus client in aibridge
coder/coder will pass in registerer, wrap with labels / name prefix

i.e.
aibridge_interception_duration
-> coderd_aibridge_interception_duration

dannykopping:
Completed, yes
In our dashboard, no
Would you mind raising a separate issue in coder/internal for that pls?

PR #62, opened 2025-11-21
feat: expose prometheus metrics

dannykopping:
Closes https://github.com/coder/aibridge/issues/25
Required for https://github.com/coder/coder/pull/20865

Metrics are updated abstractly, no provider-specific logic: all updates happen in the recorder/interceptor layers.

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -90,17 +90,17 @@ func NewMetrics(reg prometheus.Registerer) *Metrics {
 
 		// Tool-related metrics.
 
-		// Pessimistic cardinality: 2 providers, 5 models, 3 servers, 30 tools = up to 900 PER INITIATOR.
+		// Pessimistic cardinality: 2 providers, 5 models, 3 servers, 30 tools = up to 900.
 		InjectedToolUseCount: promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
 			Subsystem: "injected_tool_invocations",
 			Name:      "total",
 			Help:      "The number of times an injected MCP tool was invoked by aibridge.",
-		}, append(baseLabels, "server", "name", "initiator_id")),
-		// Pessimistic cardinality: 2 providers, 5 models, 30 tools = up to 300 PER INITIATOR.
+		}, append(baseLabels, "server", "name")),
+		// Pessimistic cardinality: 2 providers, 5 models, 30 tools = up to 300.
 		NonInjectedToolUseCount: promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
 			Subsystem: "non_injected_tool_selections",
 			Name:      "total",
 			Help:      "The number of times an AI model selected a tool to be invoked by the client.",
-		}, append(baseLabels, "name", "initiator_id")),
+		}, append(baseLabels, "name")),
 	}
 }
--- a/metrics_integration_test.go
+++ b/metrics_integration_test.go
@@ -208,7 +208,7 @@ func TestMetrics_NonInjectedToolUseCount(t *testing.T) {
 	_, _ = io.ReadAll(resp.Body)
 
 	count := promtest.ToFloat64(metrics.NonInjectedToolUseCount.WithLabelValues(
-		aibridge.ProviderOpenAI, "gpt-4.1", "read_file", userID))
+		aibridge.ProviderOpenAI, "gpt-4.1", "read_file"))
 	require.Equal(t, 1.0, count)
 }
 
@@ -268,7 +268,7 @@ func TestMetrics_InjectedToolUseCount(t *testing.T) {
 	actualServerURL := *recorder.toolUsages[0].ServerURL
 
 	count := promtest.ToFloat64(metrics.InjectedToolUseCount.WithLabelValues(
-		aibridge.ProviderAnthropic, "claude-sonnet-4-20250514", actualServerURL, mockToolName, userID))
+		aibridge.ProviderAnthropic, "claude-sonnet-4-20250514", actualServerURL, mockToolName))
 	require.Equal(t, 1.0, count)
 }
 
--- a/recorder.go
+++ b/recorder.go
@@ -215,9 +215,9 @@ func (a *AsyncRecorder) RecordToolUsage(_ context.Context, req *ToolUsageRecord)
 				if req.ServerURL != nil {
 					srvURL = *req.ServerURL
 				}
-				a.metrics.InjectedToolUseCount.WithLabelValues(a.provider, a.model, srvURL, req.Tool, a.initiatorID).Add(1)
+				a.metrics.InjectedToolUseCount.WithLabelValues(a.provider, a.model, srvURL, req.Tool).Add(1)
 			} else {
-				a.metrics.NonInjectedToolUseCount.WithLabelValues(a.provider, a.model, req.Tool, a.initiatorID).Add(1)
+				a.metrics.NonInjectedToolUseCount.WithLabelValues(a.provider, a.model, req.Tool).Add(1)
 			}
 		}
 	}()


metrics.go
```diff
@@ -0,0 +1,106 @@
...
+			Subsystem: "token_usage",
+			Name:      "total",
+			Help:      "The number of tokens used by intercepted requests.",
+		}, append(baseLabels, "type", "initiator_id")),
+
+		// Tool-related metrics.
+
+		// Estimated cardinality: 2 providers, 5 models, 3 servers, 30 tools = up to 900 PER INITIATOR.
+		InjectedToolUseCount: promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
+			Subsystem: "injected_tool_usage",
```
mtojek: re usage: `injected_tool_calls_total`?

PS. It might be worth cross-checking these names with [Prometheus naming rules](https://prometheus.io/docs/practices/naming/).

api.go
```diff
@@ -22,11 +22,12 @@ type InterceptionRecordEnded struct {
...
 type TokenUsageRecord struct {
-	InterceptionID string
-	MsgID          string
-	Input, Output  int64
-	Metadata       Metadata
-	CreatedAt      time.Time
+	InterceptionID  string
+	MsgID           string
+	Input, Output   int64
+	ExtraTokenTypes map[string]int64
```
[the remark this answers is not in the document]
mtojek: > Previously they were all being set on Metadata, but dealing with map[string]any is more tricky than a map[string]int64.

maybe drop a comment in the source code in case somebody wants to refactor that part.

bridge.go
```diff
@@ -47,20 +47,20 @@ var _ http.Handler = &RequestBridge{}
 // A [Recorder] is also required to record prompt, tool, and token use.
 //
 // mcpProxy will be closed when the [RequestBridge] is closed.
-func NewRequestBridge(ctx context.Context, providers []Provider, logger slog.Logger, recorder Recorder, mcpProxy mcp.ServerProxier) (*RequestBridge, error) {
+func NewRequestBridge(ctx context.Context, providers []Provider, recorder Recorder, mcpProxy mcp.ServerProxier, metrics *Metrics, logger slog.Logger) (*RequestBridge, error) {
```
mtojek: curious if we can carry `Metrics` together with context

btw. don't you have a feeling that there are too many params now?

intercept_openai_chat_streaming.go
```diff
@@ -161,7 +165,7 @@ func (i *OpenAIStreamingChatInterception) ProcessRequest(w http.ResponseWriter,
 				MsgID:          processor.getMsgID(),
 				Input:          calculateActualInputTokenUsage(lastUsage),
 				Output:         lastUsage.CompletionTokens,
-				Metadata: Metadata{
+				ExtraTokenTypes: map[string]int64{
```
mtojek: Actually, I'm not convinced about the property name indicating that these metrics are extra. IMHO let's preserve the original name, so `ProviderTokenDetails`, but I will not insist on renaming 👍

mtojek: BTW is there any generic method to convert ALL metrics from TokenDetails, or do we not want to do this?

interception.go
```diff
@@ -64,21 +77,37 @@ func newInterceptionProcessor(p Provider, logger slog.Logger, recorder Recorder,
...
-			slog.F("route", r.URL.Path),
+			slog.F("route", route),
 			slog.F("provider", p.Name()),
 			slog.F("interception_id", interceptor.ID()),
 			slog.F("user_agent", r.UserAgent()),
+			slog.F("streaming", interceptor.Streaming()),
 		)
 
 		log.Debug(r.Context(), "interception started")
+		if metrics != nil {
```
mtojek: ok, it is confusing to see `if metrics != nil` multiple times, should we do something about it?

metrics.go
```diff
@@ -0,0 +1,106 @@
...
+			Help:      "The count of intercepted requests.",
+		}, append(baseLabels, "status", "route", "method", "initiator_id")),
+		// Estimated cardinality: 2 providers, 5 models, 2 routes = up to 20.
+		// NOTE: route is not unbounded because this is only for intercepted routes.
+		InterceptionsInflight: promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
+			Subsystem: "interceptions",
+			Name:      "inflight",
+			Help:      "The number of intercepted requests which are being processed.",
+		}, append(baseLabels, "route")),
+		// Estimated cardinality: 2 providers, 5 models, 7 buckets + 3 extra series (count, sum, +Inf) = up to 100.
```
mtojek: Do we expect the cardinality to increase in the future?

intercept_openai_chat_streaming.go
dannykopping: They're "extra" because they're over and above the common input/output tokens. I'd like to keep this naming.

dannykopping: Not sure I understand the question. Can you rephrase?

interception.go
dannykopping: Metrics are optional and has to be a pointer type. Could make this an interface type but it seems unnecessary for now. I agree it ain't pretty.

metrics.go
dannykopping: What do you mean? Add more labels?

metrics.go
```diff
@@ -0,0 +1,106 @@
...
+		InterceptionDuration: promauto.With(reg).NewHistogramVec(prometheus.HistogramOpts{
+			Subsystem: "interceptions",
+			Name:      "duration",
+			Help: "The total duration of intercepted requests. " +
+				"The majority of this time will be the upstream processing of the request. " +
+				"aibridge has no control over upstream processing time, so it's just an illustrative metric.",
+			// TODO: add docs around determining aibridge's *own* latency with distributed traces
+			//       once https://github.com/coder/aibridge/issues/26 lands.
+			Buckets: []float64{0.5, 2, 5, 15, 30, 60, 120},
```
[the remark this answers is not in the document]
dannykopping: > is it worth adding < 0.5 buckets?

Don't think so; in my experience the APIs don't respond faster than that, generally.

> nit2: buckets > 120s?

Automatically covered by the `+Inf` bucket in prometheus histograms.

metrics.go
```diff
@@ -0,0 +1,106 @@
...
+		}, []string{"provider", "route", "method"}),
+
+		// Prompt-related metrics.
+
+		// Estimated cardinality: 2 providers, 5 models = up to 10 PER INITIATOR.
+		PromptCount: promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
+			Subsystem: "prompts",
+			Name:      "total",
+			Help:      "The number of prompts issued by users (initiators).",
+		}, append(baseLabels, "initiator_id")),
```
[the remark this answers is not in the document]
dannykopping: Operators will need to know if one user (or, by extension, agent) is overloading the system.

metrics.go
dannykopping: OK cool. I included "usage" to align with our data model, but it's probably more clear as `injected_tool_calls_total`.

What specifically do you think doesn't align with the naming rules?

metrics_integration_test.go
```diff
@@ -0,0 +1,292 @@
...
+			defer resp.Body.Close()
+			_, _ = io.ReadAll(resp.Body)
+		}
+	}()
+
+	// Wait until request is detected as inflight.
+	require.Eventually(t, func() bool {
+		return promtest.ToFloat64(
+			metrics.InterceptionsInflight.WithLabelValues(aibridge.ProviderAnthropic, "claude-sonnet-4-0", "/v1/messages"),
+		) == 1
```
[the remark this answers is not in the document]
dannykopping: Not flaky, no. The synchronisation is carefully controlled.

First the client makes a concurrenct request to the API. The API blocks until the metric shows the number of inflight requests = 1. Then it unblocks the API which completes the request.

We need `require.Eventually` because the request happens concurrently & we don't have insight from this abstraction as to when the metric will be updated.

dannykopping:
@mtojek @pawbana I removed the initiator labels on the tool-related metrics; this significantly reduces the worst-case cardinality. Please take another look.

intercept_openai_chat_streaming.go
mtojek: in pseudo-code:

`json(lastUsage.PromptTokensDetails).unmarshalToMapStringInt64()`

metrics.go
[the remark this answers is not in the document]
dannykopping: `initiator_id` == `users.id`

intercept_openai_chat_streaming.go
mtojek: We can automatically bring any new metrics an AI provider emits.

metrics.go
```diff
@@ -0,0 +1,106 @@
...
+		// NOTE: route is not unbounded because PassthroughRoutes (see provider.go) is a static list.
+		PassthroughCount: promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
+			Subsystem: "passthrough",
+			Name:      "total",
+			Help:      "The count of requests which were not intercepted but passed through to the upstream.",
+		}, []string{"provider", "route", "method"}),
+
+		// Prompt-related metrics.
+
+		// Estimated cardinality: 2 providers, 5 models = up to 10 PER INITIATOR.
```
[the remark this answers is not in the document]
mtojek: I'm asking you :) 

> ... and how many initiators?

PR #63, opened 2025-11-25
feat: add interception tracing

pawbana:
Adds tracing for interceptions, tool calls and recording.

Example interception trace:

image

PR #72, opened 2025-12-05
fix: openai blocking requests do not call injected tools correctly

dannykopping:
In https://github.com/coder/aibridge/pull/63 @pawbana uncovered a latent bug in the OpenAI non-streaming interception handler. When injected tools are invoked, arguments are not being supplied correctly to the MCP invocation. The JSON unmarshaling was silently failing, and the code in general was brittle.

This also exposed a shortcoming in the tests whereby we'd just validate that the injected tool invocation was _recorded_ (along with its args) but not that the actual MCP tool call took place! I've added a `callAccumulator` type which will track these invocations for testing.

### DIFF
--- a/bridge_integration_test.go
+++ b/bridge_integration_test.go
@@ -1270,10 +1270,10 @@ func TestStableRequestEncoding(t *testing.T) {
 			t.Cleanup(cancel)
 
 			// Setup MCP tools.
-			tools, _ := setupMCPServerProxiesForTest(t)
+			mcpProxiers, _ := setupMCPServerProxiesForTest(t)
 
 			// Configure the bridge with injected tools.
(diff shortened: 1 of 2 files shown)


SasSwart (approved):
Looks good!

I'm not sure what

```go
			var (
				args map[string]string
				buf  bytes.Buffer
			)
			_ = json.NewEncoder(&buf).Encode(tc.Function.Arguments)
			_ = json.NewDecoder(&buf).Decode(&args)
```

was intended for. As far as I can tell it converted from a formal type to just a map, but then never used that map.

dannykopping:
> but then never used that map.

It was used:

```
res, err := tool.Call(ctx, args)
```

Thanks for the review 👍

PR #75, opened 2025-12-11
feat: add circuit breaker for upstream provider overload protection

kacpersaw:
## Summary

Implement circuit breakers for upstream AI providers that detect rate limiting and overload conditions (429/503/529 status codes) and temporarily stop sending requests when providers are overloaded.

This completes the overload protection story by adding the aibridge-specific component that couldn't be implemented as generic HTTP middleware in coderd (since it requires understanding upstream provider responses).

## Key Features

- **Per-endpoint circuit breakers** - Separate circuit breakers for each endpoint within each provider (e.g., `/v1/messages` and `/v1/chat/completions` are isolated), preventing one endpoint's issues from affecting others
- **Configurable parameters** - Failure threshold, interval for clearing counts, timeout before half-open, max requests in half-open state
- **Customizable failure detection** - Pluggable `IsFailure` function to determine which status codes trigger the circuit
- **Built on sony/gobreaker** - Uses the battle-tested `github.com/sony/gobreaker/v2` library
- **Prometheus metrics** - State gauge, trips counter, rejects counter with provider/endpoint labels
- **Thread-safe** - Uses `sync.Map` for concurrent access to per-endpoint breakers
- **Opt-in per provider** - Configured via `CircuitBreakerConfig` in provider config; nil means no protection

## Circuit Breaker States

```
┌─────────┐  failures >= threshold  ┌─────────┐
│ CLOSED  │ ──────────────────────► │  OPEN   │
│(normal) │                         │(reject) │
└─────────┘                         └────┬────┘
     ▲                                   │
     │                          cooldown expires
     │                                   │
     │  successes >= threshold     ┌─────▼─────┐
     └──────────────────────────── │ HALF-OPEN │
                                   │ (testing) │
     ┌─────────────────────────────┤           │
     │         any failure         └───────────┘
     │
     ▼
┌─────────┐
│  OPEN   │
│(reject) │
└─────────┘
```

## Status Codes That Trigger Circuit Breaker (Default)

| Code | Description |
|------|-------------|
| 429 | Too Many Requests |
| 503 | Service Unavailable |
| 529 | Anthropic Overloaded |

Other error codes (400, 401, 500, 502, etc.) do **not** trigger the circuit breaker since they indicate different issues that circuit breaking wouldn't help with. Custom `IsFailure` functions can be provided to change this behavior.

## Default Configuration

| Parameter | Default | Description |
|-----------|---------|-------------|
| `FailureThreshold` | `5` | Consecutive failures needed to trip circuit |
| `Interval` | `10s` | Cyclic period for clearing internal counts in closed state |
| `Timeout` | `30s` | How long circuit stays open before transitioning to half-open |
| `MaxRequests` | `3` | Requests allowed through in half-open state |

## Prometheus Metrics

| Metric | Labels | Description |
|--------|--------|-------------|
| `circuit_breaker_state` | `provider`, `endpoint` | Current state (0=closed, 0.5=half-open, 1=open) |
| `circuit_breaker_trips_total` | `provider`, `endpoint` | Total times circuit has tripped open |
| `circuit_breaker_rejects_total` | `provider`, `endpoint` | Requests rejected due to open circuit |

## Files Changed

- `circuit_breaker.go` - Core implementation: `CircuitBreakerConfig`, `ProviderCircuitBreakers`, `CircuitBreakerMiddleware`
- `circuit_breaker_test.go` - Unit tests for middleware, failure detection, state transitions
- `circuit_breaker_integration_test.go` - End-to-end tests with mock upstream for Anthropic and OpenAI providers
- `bridge.go` - Wire up circuit breakers per provider with metrics and logging callbacks
- `provider.go` - Add `CircuitBreakerConfig()` method to `Provider` interface
- `provider_anthropic.go` - Implement `CircuitBreakerConfig()` for Anthropic
- `provider_openai.go` - Implement `CircuitBreakerConfig()` for OpenAI
- `config.go` - Add `CircuitBreaker` field to `ProviderConfig`
- `metrics.go` - Add Prometheus metrics for circuit breaker state, trips, and rejects

## Usage

```
// Enable circuit breaker for a provider
provider := aibridge.NewOpenAIProvider(aibridge.OpenAIConfig{
    BaseURL: "https://api.openai.com/v1/",
    Key:     "sk-...",
    CircuitBreaker: &aibridge.CircuitBreakerConfig{
        FailureThreshold: 5,
        Interval:         10 * time.Second,
        Timeout:          30 * time.Second,
        MaxRequests:      3,
    },
})

// Or use defaults
provider := aibridge.NewAnthropicProvider(aibridge.AnthropicConfig{
    CircuitBreaker: func() *aibridge.CircuitBreakerConfig {
        cfg := aibridge.DefaultCircuitBreakerConfig()
        return &cfg
    }(),
}, nil)
```

## Testing

All tests pass:
```
go test -count=1 -short ./...
go vet ./...
go build ./...
```

## Related

- Issue: https://github.com/coder/internal/issues/1153
- Rate limiting/concurrency PR: https://github.com/coder/coder/pull/21161

### DIFF
--- a/provider/anthropic.go
+++ b/provider/anthropic.go
@@ -32,7 +32,7 @@ var anthropicOpenErrorResponse = func() []byte {
 }
 
 var anthropicIsFailure = func(statusCode int) bool {
-    // https://platform.claude.com/docs/en/api/errors
+	// https://platform.claude.com/docs/en/api/errors
 	if statusCode == 529 {
 		return true
 	}


circuit_breaker.go
```diff
@@ -0,0 +1,349 @@
+package aibridge
```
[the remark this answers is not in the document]
pawbana: https://pkg.go.dev/github.com/go-kit/kit/circuitbreaker reuses implementations.
https://github.com/sony/gobreaker looks to be most maintained from the 3 implementations listed in go-kit. Maybe it would be better to use it directly.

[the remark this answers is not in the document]
dannykopping: > https://pkg.go.dev/github.com/go-kit/kit/circuitbreaker reuses implementations.

Whoops, yes you're right.

circuit_breaker.go
```diff
@@ -0,0 +1,131 @@
...
+func NewCircuitBreakers(config CircuitBreakerConfig, onChange func(name string, from, to gobreaker.State)) *CircuitBreakers {
+	return &CircuitBreakers{
+		config:   config,
+		onChange: onChange,
+	}
+}
+
+// Allow checks if a request to provider/endpoint should be allowed.
+func (c *CircuitBreakers) Allow(provider, endpoint string) bool {
+	if !c.config.Enabled {
```
dannykopping: If it's disabled we should create a `NoopCircuitBreaker` or something instead.

bridge.go
```diff
@@ -49,12 +54,32 @@ var _ http.Handler = &RequestBridge{}
 //
 // mcpProxy will be closed when the [RequestBridge] is closed.
 func NewRequestBridge(ctx context.Context, providers []Provider, recorder Recorder, mcpProxy mcp.ServerProxier, logger slog.Logger, metrics *Metrics, tracer trace.Tracer) (*RequestBridge, error) {
+	return NewRequestBridgeWithCircuitBreaker(ctx, providers, recorder, mcpProxy, logger, metrics, tracer, DefaultCircuitBreakerConfig())
```
dannykopping: This doesn't give us any mechanism to configure the CB going forward.
In fact, I don't think a global CB is the right strategy here. Having per-provider CBs is probably more future-proof since different providers will exhibit different behaviours.

> that couldn't be implemented as generic HTTP middleware in coderd (since it requires understanding upstream provider responses).

This would also solve this problem, since you could have provider-specific CBs.

metrics.go
```diff
@@ -102,5 +107,26 @@ func NewMetrics(reg prometheus.Registerer) *Metrics {
 			Name:      "total",
 			Help:      "The number of times an AI model selected a tool to be invoked by the client.",
 		}, append(baseLabels, "name")),
+
+		// Circuit breaker metrics.
+
+		// Pessimistic cardinality: 2 providers, 5 endpoints = up to 10.
```
dannykopping: I like this addition 👍 metrics will help a lot

circuit_breaker_test.go
```diff
@@ -0,0 +1,242 @@
+package aibridge
```
dannykopping: These tests seem to be testing the library, not the _observed behaviour_.

What I'd expect to see from a test is a mock upstream HTTP server returning errors which results in a response indicating a circuit-breaker has tripped.

bridge.go
```diff
@@ -49,12 +54,35 @@ var _ http.Handler = &RequestBridge{}
...
+	}
+	cbs := NewCircuitBreakers(cbConfigs, onChange)
+
 	for _, provider := range providers {
 		// Add the known provider-specific routes which are bridged (i.e. intercepted and augmented).
 		for _, path := range provider.BridgedRoutes() {
-			mux.HandleFunc(path, newInterceptionProcessor(provider, recorder, mcpProxy, logger, metrics, tracer))
+			handler := newInterceptionProcessor(provider, recorder, mcpProxy, logger, metrics, tracer)
+			// Wrap with circuit breaker middleware if configured for this provider
+			handler = CircuitBreakerMiddleware(cbs, metrics, provider.Name())(handler).ServeHTTP
```
pawbana: If `CircuitBreakerConfig` would be part of `Provider` struct then we would not need to pass all cbs for each `CircuitBreakerMiddleware`.

circuit_breaker.go
```diff
@@ -0,0 +1,200 @@
...
+	"github.com/sony/gobreaker/v2"
+)
+
+// CircuitBreakerConfig holds configuration for circuit breakers.
+// Fields match gobreaker.Settings for clarity.
+type CircuitBreakerConfig struct {
+	// MaxRequests is the maximum number of requests allowed in half-open state.
+	MaxRequests uint32
+	// Interval is the cyclic period of the closed state for clearing internal counts.
+	Interval time.Duration
```
pawbana: nits: not sure if better names but maybe `Interval` -> `Window`  `Timeout` -> `OpenTimeout`

circuit_breaker.go
```diff
@@ -0,0 +1,188 @@
...
+
+			_, err := cb.Execute(func() (struct{}, error) {
+				next.ServeHTTP(sw, r)
+				if cbs.config.IsFailure(sw.statusCode) {
+					return struct{}{}, fmt.Errorf("upstream error: %d", sw.statusCode)
+				}
+				return struct{}{}, nil
+			})
+
+			if err != nil && (errors.Is(err, gobreaker.ErrOpenState) || errors.Is(err, gobreaker.ErrTooManyRequests)) {
```
pawbana: nit: no need to check `err != nil`. `errors.Is` handles nils.

circuit_breaker_test.go
```diff
@@ -0,0 +1,259 @@
...
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/sony/gobreaker/v2"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCircuitBreakerMiddleware_TripsOnUpstreamErrors(t *testing.T) {
```
pawbana: This test seem to mostly test library not aibridge code, maybe verify metrics here. In general metric values should be tested somewhere.

Would be nice to have more of "integration" test that would check handlers in `NewRequestBridge` + provider configuration.

pawbana (approved):

bridge.go
```diff
@@ -48,13 +48,38 @@ var _ http.Handler = &RequestBridge{}
...
+		cfg := provider.CircuitBreakerConfig()
+		providerName := provider.Name()
+		onChange := func(endpoint string, from, to gobreaker.State) {
+			logger.Info(context.Background(), "circuit breaker state change",
+				slog.F("provider", providerName),
+				slog.F("endpoint", endpoint),
+				slog.F("from", from.String()),
+				slog.F("to", to.String()),
+			)
+			if cfg != nil && metrics != nil {
```
ssncferreira: Do we need to check if `cfg != nil`?
```suggestion
			if metrics != nil {
```

circuit_breaker_integration_test.go
```diff
@@ -0,0 +1,296 @@
...
+			}
+			mockSrv.Start()
+
+			makeRequest := func() *http.Response {
+				req, _ := http.NewRequest("POST", mockSrv.URL+"/"+tc.providerName+tc.endpoint, strings.NewReader(tc.requestBody))
+				req.Header.Set("Content-Type", "application/json")
+				tc.setupHeaders(req)
+				resp, err := http.DefaultClient.Do(req)
+				require.NoError(t, err)
+				_, _ = io.ReadAll(resp.Body)
```
ssncferreira: nit:
```suggestion
				_, err = io.ReadAll(resp.Body)
				require.NoError(t, err)
```

circuit_breaker_integration_test.go
```diff
@@ -0,0 +1,296 @@
...
+
+	makeRequest := func() *http.Response {
+		req, _ := http.NewRequest("POST", mockSrv.URL+"/openai/v1/chat/completions",
+			strings.NewReader(`{"model":"gpt-4o","messages":[{"role":"user","content":"hi"}]}`))
+		req.Header.Set("Content-Type", "application/json")
+		req.Header.Set("Authorization", "Bearer test-key")
+		resp, err := http.DefaultClient.Do(req)
+		require.NoError(t, err)
+		_, _ = io.ReadAll(resp.Body)
+		resp.Body.Close()
```
ssncferreira: Same error checks as previous comments

circuit_breaker_integration_test.go
```diff
@@ -0,0 +1,296 @@
...
+			assert.Equal(t, 1.0, trips, "CircuitBreakerTrips should be 1")
+
+			state := promtest.ToFloat64(metrics.CircuitBreakerState.WithLabelValues(tc.providerName, tc.endpoint))
+			assert.Equal(t, 1.0, state, "CircuitBreakerState should be 1 (open)")
+
+			rejects := promtest.ToFloat64(metrics.CircuitBreakerRejects.WithLabelValues(tc.providerName, tc.endpoint))
+			assert.Equal(t, 1.0, rejects, "CircuitBreakerRejects should be 1")
+
+			// Phase 3: Wait for timeout to transition to half-open
+			time.Sleep(cbConfig.Timeout + 10*time.Millisecond)
```
ssncferreira: This `time.Sleep` approach for testing state transitions is flake-prone 😕 Could we use `quartz` to adjust the time and make the tests deterministic?

kacpersaw: To use `quartz` we would need to fork gobreaker to add clock injection. Alternatively we can use `testing/synctest`(which is experimental) which will create a fake time bubble. (https://go.dev/blog/synctest#testing-time)

wdyt?

ssncferreira: Oh, that is unfortunate 😕 thank you for researching that. I think we can leave as is now, and if needed, address it in a separate PR 👍

ssncferreira (approved):
LGTM 👍 Thanks for addressing the comments!

bridge.go
```diff
@@ -59,22 +60,47 @@ const recordingTimeout = time.Second * 5
...
 
 		// Any requests which passthrough to this will be reverse-proxied to the upstream.
 		//
 		// We have to whitelist the known-safe routes because an API key with elevated privileges (i.e. admin) might be
 		// configured, so we should just reverse-proxy known-safe routes.
-		ftr := newPassthroughRouter(provider, logger.Named(fmt.Sprintf("passthrough.%s", provider.Name())), m, tracer)
-		for _, path := range provider.PassthroughRoutes() {
-			prefix := fmt.Sprintf("/%s", provider.Name())
+		ftr := newPassthroughRouter(prov, logger.Named(fmt.Sprintf("passthrough.%s", prov.Name())), m, tracer)
+		for _, path := range prov.PassthroughRoutes() {
```
ssncferreira: IIUC, we are only adding a circuit breaker to bridged routes, right? Passthrough routes don't get circuit breaker protection, is this intentional?

provider/anthropic.go
```diff
@@ -102,3 +103,10 @@ func (p *Anthropic) InjectAuthHeader(headers *http.Header) {
 
 	headers.Set(p.AuthHeader(), p.cfg.Key)
 }
+
+func (p *Anthropic) CircuitBreakerConfig() *circuitbreaker.Config {
+	if p.cfg.CircuitBreaker != nil && p.cfg.CircuitBreaker.IsFailure == nil {
+		p.cfg.CircuitBreaker.IsFailure = circuitbreaker.AnthropicIsFailure
```
dannykopping: This is pretty confusing. Why does `AnthropicIsFailure` exist in the `circuitbreaker` pkg? It should exist here.

circuitbreaker/circuitbreaker.go
```diff
@@ -0,0 +1,203 @@
...
+	MaxRequests uint32
+	// Interval is the cyclic period of the closed state for clearing internal counts.
+	Interval time.Duration
+	// Timeout is how long the circuit stays open before transitioning to half-open.
+	Timeout time.Duration
+	// FailureThreshold is the number of consecutive failures that triggers the circuit to open.
+	FailureThreshold uint32
+	// IsFailure determines if a status code should count as a failure.
+	// If nil, defaults to 429 and 503 (see DefaultIsFailure).
+	// Anthropic provider uses AnthropicIsFailure which also includes 529.
```
dannykopping: Nit: keep implementation details out of comments; they go stale quickly.

```suggestion
	// If nil, defaults to 429 and 503 (see DefaultIsFailure).
	// Anthropic provider uses AnthropicIsFailure which also includes 529.
```

circuitbreaker/circuitbreaker.go
```diff
@@ -0,0 +1,203 @@
...
+		IsFailure:        DefaultIsFailure,
+	}
+}
+
+// DefaultIsFailure returns true for standard HTTP status codes that typically
+// indicate upstream overload: 429 (Too Many Requests) and 503 (Service Unavailable).
+func DefaultIsFailure(statusCode int) bool {
+	switch statusCode {
+	case http.StatusTooManyRequests, // 429
+		http.StatusServiceUnavailable: // 503
```
dannykopping: We should include 504 as well.

provider/anthropic.go
dannykopping: Also, this is mutating in a getter.
Can we rather move this to `NewAnthropic()`?

circuitbreaker/circuitbreaker.go
```diff
@@ -0,0 +1,203 @@
...
+			})
+
+			if errors.Is(err, gobreaker.ErrOpenState) || errors.Is(err, gobreaker.ErrTooManyRequests) {
+				if m != nil {
+					m.CircuitBreakerRejects.WithLabelValues(cbs.provider, endpoint).Inc()
+				}
+				w.Header().Set("Content-Type", "application/json")
+				w.Header().Set("Retry-After", strconv.FormatInt(int64(cbs.config.Timeout.Seconds()), 10))
+				w.WriteHeader(http.StatusServiceUnavailable)
+				w.Write([]byte(`{"type":"error","error":{"type":"circuit_breaker_open","message":"circuit breaker is open"}}`))
```
dannykopping: This error format needs to be specific to each provider.

circuitbreaker/circuitbreaker.go
```diff
@@ -7,85 +7,64 @@ import (
...
+	}
+	return DefaultIsFailure(statusCode)
+}
+
+// openErrorResponse returns the error response body when the circuit is open.
+func (p *ProviderCircuitBreakers) openErrorResponse() []byte {
+	if p.config.OpenErrorResponse != nil {
+		return p.config.OpenErrorResponse()
+	}
+	return []byte(`{"error":"circuit breaker is open"}`)
```
dannykopping: Each provider has its own error type which can be marshaled the way they expect to return errors. We should use that.
See `intercept/messages/base.go -> newErrorResponse()`  for example.

provider/anthropic.go
```diff
@@ -27,13 +27,28 @@ type Anthropic struct {
 
 const routeMessages = "/anthropic/v1/messages" // https://docs.anthropic.com/en/api/messages
 
+var anthropicOpenErrorResponse = func() []byte {
+	return []byte(`{"type":"error","error":{"type":"overloaded_error","message":"circuit breaker is open"}}`)
+}
+
+var anthropicIsFailure = func(statusCode int) bool {
+	if statusCode == 529 {
```
dannykopping:
```suggestion
    // https://platform.claude.com/docs/en/api/errors
	if statusCode == 529 {
```

PR #78, opened 2025-12-15
fix: correctly set `disable_parallel_tool_use`

dannykopping:
Relates to #2.

Currently we're indiscriminately setting `disable_parallel_tool_use` in the `any` type: this is a correctness bug.
We should **only** set this field for the type which is set.

https://platform.claude.com/docs/en/agents-and-tools/tool-use/implement-tool-use#parallel-tool-use

There are 4 types: `auto`, `any`, `tool`, `none`. `none` is irrelevant (can't parallelize tool calls which don't occur).

We should check which type is used (or set `auto` if none is provided) and set the field on that.

---

I noticed this today when using Claude Code v2.0.69.

The request did not contain a `tool_choice` field (behavioural change), so I got this error:

```json
{
    "error": {
        "message": "Thinking may not be enabled when tool_choice forces tool use.",
        "type": "invalid_request_error"
    },
    "request_id": "",
    "type": "error"
}
```

### DIFF
--- a/bridge_integration_test.go
+++ b/bridge_integration_test.go
@@ -22,6 +22,7 @@ import (
 	"cdr.dev/slog/sloggers/slogtest"
 	"github.com/anthropics/anthropic-sdk-go"
 	"github.com/anthropics/anthropic-sdk-go/packages/ssestream"
+	"github.com/anthropics/anthropic-sdk-go/shared/constant"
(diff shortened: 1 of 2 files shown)


ssncferreira (approved):
LGTM, just a small nit 👍

PR #96, opened 2025-12-23
fix: only disable parallel tool use when tools are injected

dannykopping:
This change adds an early return in injectTools() when there are no
tools to inject from MCP, preventing unnecessary modification of the
tool_choice parameter. Previously, disable_parallel_tool_use was set
even when no tools were injected, which was incorrect behaviour.


ssncferreira (approved):
LGTM, just some minor comments 🍰

PR #99, opened 2025-12-23
chore: rearrange code into packages

pawbana:
# Refactor: Reorganize code into packages

Adds following packages
- config: contains config structs and some constants
- context: actor related context functions
- intercept:
  - chatcompletions: OpenAI chat completions interceptors
  - eventstream: common even stream logic beween chatcompletions/messages packages
  - messages: Anthrpic messages interceptors
- provider: OpenAI/Antropic providers
- recorder

The main API surface is preserved through type aliases and function wrappers in api.go, ensuring backward compatibility while improving the internal code organization.

### DIFF
--- a/sse_parser.go
+++ b/sse_parser.go
@@ -9,33 +9,33 @@ import (
 )
 
 const (
-	sseEventTypeMessage = "message"
-	sseEventTypeError   = "error"
-	sseEventTypePing    = "ping"
+	SSEEventTypeMessage = "message"
+	SSEEventTypeError   = "error"
+	SSEEventTypePing    = "ping"
 )
 
-type sseEvent struct {
+type SSEEvent struct {
 	Type  string
 	Data  string
 	ID    string
 	Retry int
 }
 
-type sseParser struct {
-	events map[string][]sseEvent
+type SSEParser struct {
+	events map[string][]SSEEvent
 	mu     sync.RWMutex
 }
 
-func NewSSEParser() *sseParser {
-	return &sseParser{
-		events: make(map[string][]sseEvent),
+func NewSSEParser() *SSEParser {
+	return &SSEParser{
+		events: make(map[string][]SSEEvent),
 	}
 }
 
-func (p *sseParser) Parse(reader io.Reader) error {
+func (p *SSEParser) Parse(reader io.Reader) error {
 	scanner := bufio.NewScanner(reader)
 
-	var currentEvent sseEvent
+	var currentEvent SSEEvent
 	var dataLines []string
 
 	for scanner.Scan() {
@@ -49,7 +49,7 @@ func (p *sseParser) Parse(reader io.Reader) error {
 
 			// Default to message type if no event type specified
 			if currentEvent.Type == "" {
-				currentEvent.Type = sseEventTypeMessage
+				currentEvent.Type = SSEEventTypeMessage
 			}
 
 			// Store the event
@@ -58,7 +58,7 @@ func (p *sseParser) Parse(reader io.Reader) error {
 			p.mu.Unlock()
 
 			// Reset for next event
-			currentEvent = sseEvent{}
+			currentEvent = SSEEvent{}
 			dataLines = nil
 			continue
 		}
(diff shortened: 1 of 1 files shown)


pawbana:
* **#99** <a href="https://app.graphite.com/github/pr/coder/aibridge/99?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/aibridge/99?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

config/config.go
```diff
@@ -1,14 +1,15 @@
...
-
-type (
-	OpenAIConfig    ProviderConfig
-	AnthropicConfig ProviderConfig
+const (
+	ProviderAnthropic = "anthropic"
+	ProviderOpenAI    = "openai"
 )
 
+type AnthropicConfig struct {
```
dannykopping: Nit: since this is now in the `config` package, these structs no longer need the `Config` suffix.

example/main.go
```diff
@@ -55,23 +57,28 @@ func main() {
 	reg := prometheus.NewRegistry()
 	metrics := aibridge.NewMetrics(reg)
 
+	// Setup tracing
+	tracer := otel.GetTracerProvider().Tracer("exampleTracer")
```
dannykopping: Oh nice, thanks for updating the example.
We should probably have a CI job to validate this example compiles.

intercept/chatcompletions/base.go
```diff
@@ -0,0 +1,249 @@
+package chatcompletions
```
dannykopping: I don't really understand the package structure of `intercept`.

The high-level packages (`chatcompletions`, `messages`, `eventstream`) are all for different purposes or providers. This also means that code-reuse across different endpoint handlers for the same provider can't share code within their same package, so you'd have to export those - which would be awkward.

intercept/chatcompletions/base.go
```diff
@@ -0,0 +1,249 @@
...
+	"github.com/openai/openai-go/v2"
+	"github.com/openai/openai-go/v2/option"
+	"github.com/openai/openai-go/v2/shared"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+
+	"cdr.dev/slog"
+)
+
+type InterceptionBase struct {
```
dannykopping: We need to be _very_ deliberate about only exporting types & funcs that could conceivably be used by users of this library.

I can't see a reason why it's exported, since it only has a package-local role.

_Remember: exported types & funcs increase the API surface, which means a larger maintenance burden for us going forward._

context/context.go
```diff
@@ -0,0 +1,29 @@
+package context
```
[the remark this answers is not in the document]
pawbana: `*Config` structs are used i chatcompletions, messages and provider packages. Configs are also used in coder repo.
It would be possible to define configs in base of chatcompletions/messages packages but it seemed less clear approach then separate package to me.
Later this package should probably be removed and references from chatcompletions/messages packages should be removed.

intercept/chatcompletions/base.go
pawbana: My idea behind `intercept` package is for it to contain all interceptor implementations + common logic between interceptors. Each sub packages is specific interceptor implementation, each sub package implementing interceptor interface. It may require different structure later as this is the place where most code will probably be added, I can see it becoming too large / new "root" package.

`eventstream` is used in both `chatcompletions` and `messages` packages. `envetstream` isn't useful outside of interception context so I located in `intercept` package as common logic between interceptors. I don't see it being in separate sub-package as awkward.

intercept/messages/base.go
```diff
@@ -257,3 +267,111 @@ func (t *redirectTransport) RoundTrip(req *http.Request) (*http.Response, error)
 
 	return t.base.RoundTrip(req)
 }
+
+// AccumulateUsage accumulates usage statistics from source into dest.
+// It handles both [anthropic.Usage] and [anthropic.MessageDeltaUsage] types through [any].
+// The function uses reflection to handle the differences between the types:
+// - [anthropic.Usage] has CacheCreation field with ephemeral tokens
+// - [anthropic.MessageDeltaUsage] doesn't have CacheCreation field
+func AccumulateUsage(dest, src any) {
```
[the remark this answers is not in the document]
pawbana: Good catch, sorry about that. I'll double check if I have not leaked any more methods.

Most packages could be moved into `internal` package. I didn't want to do it to keep changes as simple as possible but now that I think about it, it should not cause to much confusion.

api.go
```diff
@@ -2,67 +2,54 @@ package aibridge
...
-	Metadata                     Metadata
-	StartedAt                    time.Time
-}
+// Type + function aliases for backwards compatibility.
+type (
+	Metrics = metrics.Metrics
+
+	Provider = provider.Provider
+
+	InterceptionRecord      = recorder.InterceptionRecord
```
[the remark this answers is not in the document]
pawbana: I added re-exporting to keep changes to minimum (also in coder repo). Maybe it is lazy of me but I think in aibridge lib case there is no that many types/functions exported so it would still be clear and easier to use if everything was exported by root package.

I think re-export approach + `internal` package would work well here but I didn't create `internal` package to keep changes to minimum.

context/context.go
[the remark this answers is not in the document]
pawbana: `ActorFromContext` is used in interceptors so it needs to be separate package to not have import cycles.

It could be part of root package if actor would be a parameter not extracted from context which makes sense as interceptor must have defined actor but would require more changes.

dannykopping (approved):
Let's merge this as-is for now. There are a few things I'm not 100% in agreement with (from a taste perspective) but I think this does move us forward, and doesn't tie our hands.

Thank you for doing this @pawbana

PR #103, opened 2025-12-30
perf: stream anthropic request bodies for bridge

dannykopping:
Stream decode anthropic requests to bridge to avoid the unnecessary allocation of a potentially large request payload.


dannykopping:
@SasSwart anything blocking this merge?

Issue #84, opened 2025-12-22
Create basic /v1/response interceptor


pawbana:
Please create basic interceptor supporting OpenAI /v1/responses API.

Basic interceptor should only record interception start and end and forward requests / responses / events.

Do not add recording of Prompt/Token/Tool usage or MCP tool injection support.

Add passthrough routes for:
* conversations: https://platform.openai.com/docs/api-reference/conversations
* other then `/create` responses API endpoints (eg. [get](https://platform.openai.com/docs/api-reference/responses/get)): https://platform.openai.com/docs/api-reference/responses

For issue to be resolved responses and conversations API should work correctly though bridge.

PR #107, opened 2026-01-08
feat: add basic responses API interceptor


pawbana:
Adds basic interceptor for openAI responses API.
For now responses API interceptor doesn't record Prompt/Token/Tool usage and doesn't support MCP tool injection.

Closes: https://github.com/coder/aibridge/issues/84

### DIFF
--- a/intercept/responses/base.go
+++ b/intercept/responses/base.go
@@ -172,13 +172,13 @@ func (r *responseCopier) copyMiddleware(req *http.Request, next option.Middlewar
 	r.responseReceived.Store(true)
 	r.responseStatus = resp.StatusCode
 	r.responseHeaders = resp.Header
-	r.responseBody = resp.Body
 	resp.Body = io.NopCloser(io.TeeReader(resp.Body, &r.buff))
+	r.responseBody = resp.Body
 	return resp, nil
 }
 
-// readAll reads all data from original request body
-// so TeeReader copies it to buffer and returns buffer contents.
+// readAll reads all data from resp.Body returned by so TeeReader
+// so it appends all read data to the buffer and returns buffer contents.
 func (r *responseCopier) readAll() ([]byte, error) {
 	if r.responseBody == nil {
 		return []byte{}, nil


pawbana:
* **#107** <a href="https://app.graphite.com/github/pr/coder/aibridge/107?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/aibridge/107?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

intercept/responses/base.go
```diff
@@ -0,0 +1,184 @@
...
+func (d *deltaBuffer) Write(p []byte) (int, error) {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+	return d.buf.Write(p)
+}
+
+func (d *deltaBuffer) Read(p []byte) (int, error) {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+	c, err := d.buf.Read(p)
```
mtojek: nit: `return d.buf.Read(p)`

responses_integration_test.go
```diff
@@ -0,0 +1,345 @@
...
+			name:      "streaming",
+			streaming: true,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			ctx, cancel := context.WithTimeout(t.Context(), time.Second*30)
```
mtojek: nit: It may be worth bringing a [similar file](https://github.com/coder/coder/blob/4bc49ed6eb8c79b6be1924477a1fcfda41e78ed1/testutil/duration.go#L12) here

responses_integration_test.go
```diff
@@ -0,0 +1,345 @@
...
+
+			var receivedRequest map[string]any
+			upstream := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				raw, err := io.ReadAll(r.Body)
+				require.NoError(t, err)
+				defer r.Body.Close()
+
+				require.NoError(t, json.Unmarshal(raw, &receivedRequest))
+
+				w.WriteHeader(http.StatusOK)
```
mtojek: nit: should we check if upstream really received the requests and replied?

intercept/responses/streaming.go
```diff
@@ -0,0 +1,103 @@
...
+		return upstreamErr
+	}
+
+	// Sometimes stream.Next() returns before respBody buffer is filled
+	lastRead, err := io.ReadAll(&respBody)
+	if err != nil {
+		i.logger.Warn(ctx, "failed to read upstream response", slog.Error(err))
+		return fmt.Errorf("failed to read upstream response: %w", err)
+	}
+	events.Send(ctx, lastRead)
```
mtojek: alternative form?

```go
	if remaining := respBody.readDelta(); len(remaining) > 0 {
		_ = events.Send(ctx, remaining)
	}
```

intercept/responses/streaming.go
```diff
@@ -0,0 +1,103 @@
...
+
+func (i *StreamingResponsesInterceptor) Streaming() bool {
+	return true
+}
+
+func (i *StreamingResponsesInterceptor) TraceAttributes(r *http.Request) []attribute.KeyValue {
+	return i.responsesInterceptionBase.baseTraceAttributes(r, true)
+}
+
+func (i *StreamingResponsesInterceptor) ProcessRequest(w http.ResponseWriter, r *http.Request) error {
```
mtojek: Ok, I have a feeling that you over-engineered this fragment. If we get rid of all cancel/defer noise, the flow should be like:

1. validateRequest
2. eventstream.NewEventStream + go events.Start
3. srv.NewStreaming
4. for stream.Next
5. handleUpstreamError
6. context.WithTimeout + events.Shutdown

As you can see I reduced all context cancellations to only the last one.

intercept/responses/blocking.go
```diff
@@ -0,0 +1,73 @@
...
+
+	w.Header().Set("Content-Type", "application/json")
+	out, err := io.ReadAll(&respBody)
+	if err != nil {
+		i.logger.Warn(ctx, "failed to read upstream response", slog.Error(err))
+		return errors.Join(fmt.Errorf("failed to read upstream response: %w", err), upstreamErr)
+	}
+
+	_, err = w.Write(out)
+	if err != nil {
```
mtojek: nit: could be inline

intercept/responses/base.go
pawbana: refactored.

intercept/responses/base.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+}
+
+func (i *responsesInterceptionBase) validateRequest(ctx context.Context, w http.ResponseWriter) error {
+	if i.req == nil {
+		return errors.New("developer error: req is nil")
+	}
+
+	if i.req.Background.Value {
+		err := fmt.Errorf("background requests are currently not supported by aibridge")
```
dannykopping: Seems unnecessary to create an error only to just read its string value out again.
If the accept header is JSON, you'll need to response with JSON as well.
I'd suggest mimicking the OpenAI error format so that clients will at least be able to render the error correctly.

intercept/eventstream/eventstream.go
```diff
@@ -209,7 +209,7 @@ func IsConnError(err error) bool {
 		return true
 	}
 
-	if errors.Is(err, syscall.ECONNRESET) || errors.Is(err, syscall.EPIPE) || errors.Is(err, net.ErrClosed) {
+	if errors.Is(err, syscall.ECONNRESET) || errors.Is(err, syscall.EPIPE) || errors.Is(err, syscall.ECONNREFUSED) || errors.Is(err, net.ErrClosed) {
```
dannykopping: What's the reason for this change?
`ECONNREFUSED` is semantically different I feel.

intercept/responses/base.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+	if !i.req.Stream {
+		opts = append(opts, option.WithRequestTimeout(requestTimeout))
+	}
+	return opts
+}
+
+// copyMiddleware copies Content-Type header, status code and
+// response body to given buffer leaving original response
+// intact/consumable for openAI SDK
```
dannykopping: Nit: comments should discuss the _why_ not the _what_, since the _what_ can change and leave the comment stale.

This func is pretty crucial, and needs a lot of description please.

intercept/responses/base.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+func (d *deltaBuffer) Write(p []byte) (int, error) {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+	return d.buf.Write(p)
+}
+
+// Reads all data from original resqusts body
+// so it is properly copied by TeeReader to buffer
```
dannykopping: This relies on an implementation detail of how its used, which is a bit awkward.

intercept/responses/base.go
```diff
@@ -0,0 +1,197 @@
...
+	http.Error(w, upstreamErr.Error(), http.StatusInternalServerError)
+	return fmt.Errorf("upstream connection error: %w", upstreamErr), true
+}
+
+// deltaBuffer is a thread safe byte buffer
+// supports reading incremental data (added after last read)
+type deltaBuffer struct {
+	mu     sync.RWMutex
+	buf    bytes.Buffer
+	closer io.ReadCloser
```
dannykopping: Nit: the names here don't really indicate their purpose.

intercept/responses/base.go
```diff
@@ -0,0 +1,197 @@
...
+	return func(req *http.Request, next option.MiddlewareNext) (*http.Response, error) {
+		resp, err := next(req)
+		if err != nil || resp == nil {
+			return resp, err
+		}
+
+		// mark that some response has been received
+		i.gotResponse.Store(true)
+
+		w.Header().Set("Content-Type", resp.Header.Get("Content-Type"))
```
dannykopping: Why is `Content-Type` special, but not the other headers?

intercept/responses/paramswrap.go
```diff
@@ -0,0 +1,26 @@
...
+// in responses.ResponseNewParams.
+type ResponsesNewParamsWrapper struct {
+	responses.ResponseNewParams
+	Stream bool `json:"stream,omitempty"`
+}
+
+func (c *ResponsesNewParamsWrapper) UnmarshalJSON(raw []byte) error {
+	err := c.ResponseNewParams.UnmarshalJSON(raw)
+	if err != nil {
+		return err
```
dannykopping: Wrap error please.

intercept/responses/base.go
```diff
@@ -0,0 +1,197 @@
...
+
+	i.logger.Warn(ctx, "openai responses API upstream error", slog.Error(upstreamErr))
+
+	if i.gotResponse.Load() {
+		// if copyMiddleware received some response forward it as is
+		return fmt.Errorf("responses API error: %w", upstreamErr), false
+	}
+
+	// if no response has been received (eg. client/connection error) return internal server error
+	http.Error(w, upstreamErr.Error(), http.StatusInternalServerError)
```
dannykopping: Body may have been written at this point; this may not be safe to do.

intercept/responses/streaming.go
```diff
@@ -0,0 +1,95 @@
...
+
+	srv := i.newResponsesService()
+	opts := i.requestOptions(&respPayload, w)
+	stream := srv.NewStreaming(ctx, i.req.ResponseNewParams, opts...)
+	defer stream.Close()
+
+	for stream.Next() {
+		if err := events.Send(ctx, respPayload.readDelta()); err != nil {
+			i.logger.Warn(ctx, "failed to relay chunk", slog.Error(err))
+			err = fmt.Errorf("relay chunk: %w", err)
```
dannykopping: `err` is assigned but never used; it's shadowed and then lost.

intercept/responses/base.go
mtojek: probably worth dropping a comment in the code

intercept/responses/base.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+// readDelta returns only the bytes appended
+// after the last readDelta call.
+func (d *deltaBuffer) readDelta() []byte {
+	d.mu.RLock()
+	defer d.mu.RUnlock()
+
+	b := bytes.Clone(d.buf.Bytes())
+	d.buf.Reset()
```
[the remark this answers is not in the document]
mtojek: Most likely that's missight from the [previous iteration](https://github.com/coder/aibridge/commit/cc095651845edb402ef5e8dd47bd9777d01710eb) 👍 

Could we use `RLock()` while `bytes.Clone`ing and then upgrade to `Lock` just to reset the buffer?

intercept/responses/base.go
```diff
@@ -0,0 +1,197 @@
...
+		}
+	}
+
+	return nil
+}
+
+func (i *responsesInterceptionBase) requestOptions(payloadBuff *deltaBuffer, w http.ResponseWriter) []option.RequestOption {
+	opts := []option.RequestOption{
+		// Sends original payload to solve json re-encoding issues
+		// eg. Codex CLI produces requests without ID set in reasoning items: https://platform.openai.com/docs/api-reference/responses/create#responses_create-input-input_item_list-item-reasoning-id
```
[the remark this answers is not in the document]
pawbana: Example (lines ~37): https://github.com/coder/aibridge/pull/109/changes#diff-9cce08021152dd899d07926e55c7e1dece38afc2492eae1694916934583b1ecb

When this request is re-encoded by SDK an `id: ""` field is added to reasoning item which causes upstream to return 400 bad request: `id` value must match some patter error.

intercept/responses/base.go
```diff
@@ -0,0 +1,197 @@
...
+
+	return responses.NewResponseService(opts...)
+}
+
+func (i *responsesInterceptionBase) ID() uuid.UUID {
+	return i.id
+}
+
+func (i *responsesInterceptionBase) Setup(logger slog.Logger, recorder recorder.Recorder, mcpProxy mcp.ServerProxier) {
+	i.logger = logger.With(slog.F("model", i.req.Model))
```
ssncferreira: Nit: 
```suggestion
	i.logger = logger.With(slog.F("model", i.model))
```

provider/openai.go
```diff
@@ -56,8 +63,10 @@ func (p *OpenAI) BridgedRoutes() []string {
 func (p *OpenAI) PassthroughRoutes() []string {
 	return []string{
 		"/v1/models",
-		"/v1/models/",   // See https://pkg.go.dev/net/http#hdr-Trailing_slash_redirection-ServeMux.
-		"/v1/responses", // TODO: support Responses API.
+		"/v1/models/",    // See https://pkg.go.dev/net/http#hdr-Trailing_slash_redirection-ServeMux.
+		"/v1/responses/", // Forwards other responses API endpoints, eg: https://platform.openai.com/docs/api-reference/responses/get
+		"/v1/conversations",
+		"/v1/conversations/",
```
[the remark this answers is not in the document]
pawbana: added comment

intercept/eventstream/eventstream.go
pawbana: Reverted

intercept/responses/streaming.go
pawbana: refactored

responses_integration_test.go
```diff
@@ -0,0 +1,528 @@
...
+	t.Parallel()
+
+	tests := []struct {
+		name                         string
+		request                      string
+		streaming                    bool
+		expectParallelToolCalls      bool
+		expectParallelToolCallsValue bool
+	}{
+		{
```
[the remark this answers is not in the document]
pawbana: Used the same logic as in chat completions: https://github.com/coder/aibridge/blob/7535a71e91a1d214a31a9b59bb810befb26141bc/intercept/chatcompletions/streaming.go#L99

responses_integration_test.go
```diff
@@ -0,0 +1,528 @@
...
+			streaming:   true,
+			statusCode:  http.StatusBadRequest,
+			contentType: "application/json",
+			body:        responsesError,
+		},
+		{
+			name:            "blocking_non_responses_error",
+			streaming:       false,
+			statusCode:      http.StatusBadGateway,
+			contentType:     "text/plain",
```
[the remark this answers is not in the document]
pawbana: I don't know. This test is not about clients using `text/plain` but about testing whatever content type is received from upstream is forwarded to the client.

intercept/responses/base.go
dannykopping: Can you explain why we can't forward all the headers?

responses_integration_test.go
```diff
@@ -0,0 +1,528 @@
...
+			resp, err := client.Do(req)
+			require.NoError(t, err)
+			defer resp.Body.Close()
+			_, err = io.ReadAll(resp.Body)
+			require.NoError(t, err)
+		})
+	}
+}
+
+// TODO set MaxRetries to speed up this test
```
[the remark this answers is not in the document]
pawbana: added comment, created https://github.com/coder/aibridge/issues/115

intercept/responses/base.go
```diff
@@ -0,0 +1,227 @@
...
+// responseCopier helper struct to send original response to the client
+type responseCopier struct {
+	buff            deltaBuffer
+	responseStatus  int
+	responseHeaders http.Header
+
+	// responseBody keeps reference to original ReadCloser.
+	// TeeReader in copyMiddleware copies read bytes from
+	// response body (read by SDK) to the buffer. In case
+	// SDK doesns't read everything readAll method reads from
```
dannykopping: "In case SDK doesns't read everything"  
  
When would this happen?

intercept/responses/base.go
```diff
@@ -0,0 +1,227 @@
...
+		// eliminating any possibility of JSON re-encoding issues.
+		option.WithMiddleware(respCopy.copyMiddleware),
+	}
+	if !i.req.Stream {
+		opts = append(opts, option.WithRequestTimeout(requestTimeout))
+	}
+	return opts
+}
+
+// responseCopier helper struct to send original response to the client
```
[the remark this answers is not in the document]
dannykopping: I'm still not entirely convinced about this design but we can revisit it later.  
For now I'd like to make sure all the reasoning is captured, including issues we previously faced with re-encoding, so when we revisit this we have all the context. You might need to describe the re-encoding issues in detail somewhere, or reference a previous issue.

intercept/responses/base.go
dannykopping: This kind of premature defensiveness is difficult for me. I'd rather we solve actual issues which are likely.
There are an infinity of defences we could put in place against the SDK's behaviour, but we have to go by the contract that they expose and trust it.

intercept/responses/base.go
```diff
@@ -0,0 +1,224 @@
...
+			i.sendCustomErr(ctx, w, http.StatusInternalServerError, err)
+			return err
+		}
+	}
+
+	return nil
+}
+
+// sendCustomErr sends custom responses.Error error to the client
+// it should only be called before any data is sent back to the client
```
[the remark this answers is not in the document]
pawbana: Created: https://github.com/coder/aibridge/issues/116

dannykopping (approved):

Issue #86, opened 2025-12-22
Add prompt recording to /v1/responses interceptor


pawbana:
Add prompt recording to /v1/responses interceptor. Both streaming and blocking.

PR #109, opened 2026-01-12
feat: add prompt recording for responses API

pawbana:
Adds prompt recording to responses interceptors.
Closes: https://github.com/coder/aibridge/issues/86

### DIFF
--- a/intercept/responses/base.go
+++ b/intercept/responses/base.go
@@ -163,7 +163,7 @@ func (i *responsesInterceptionBase) lastUserPrompt() (string, error) {
 		return i.req.Input.OfString.Value, nil
 	}
 
-	// fallback to parsing original bytes since golang SDK doensn't properly decode 'Input' field.
+	// Fallback to parsing original bytes since golang SDK doesn't properly decode 'Input' field.
 	// If 'type' field of input item is not set it will be omitted from 'Input.OfInputItemList'
 	// It is an optional field according to API: https://platform.openai.com/docs/api-reference/responses/create#responses_create-input-input_item_list-input_message
 	// example: fixtures/openai/responses/blocking/builtin_tool.txtar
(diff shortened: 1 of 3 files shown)


bridge_integration_test.go
```diff
@@ -44,35 +44,7 @@ import (
...
-	//go:embed fixtures/openai/chatcompletions/fallthrough.txtar
-	oaiFallthrough []byte
-	//go:embed fixtures/openai/chatcompletions/stream_error.txtar
-	oaiMidStreamErr []byte
-	//go:embed fixtures/openai/chatcompletions/non_stream_error.txtar
-	oaiNonStreamErr []byte
-
-	testTracer = otel.Tracer("forTesting")
-)
+var testTracer = otel.Tracer("forTesting")
```
dannykopping: I really like this change, but it should be in its own PR.

It muddies the intent of the actual PR.

intercept/responses/blocking.go
```diff
@@ -47,22 +47,29 @@ func (i *BlockingResponsesInterceptor) ProcessRequest(w http.ResponseWriter, r *
...
+	response, upstreamErr := srv.New(ctx, i.req.ResponseNewParams, opts...)
 
 	upstreamErr, failFast := i.handleUpstreamError(ctx, upstreamErr, w)
 	if failFast {
 		return upstreamErr
 	}
 
-	if err := respPayload.drainCloser(); err != nil {
+	// response could be nil eg. fixtures/openai/responses/blocking/wrong_response_format.txtar
+	if response != nil {
```
dannykopping: I assume it could only be `nil` when an error is present, right?

intercept/responses/streaming.go
```diff
@@ -65,15 +66,22 @@ func (i *StreamingResponsesInterceptor) ProcessRequest(w http.ResponseWriter, r
...
+	opts := i.requestOptions(&responsePayload, w)
 	stream := srv.NewStreaming(ctx, i.req.ResponseNewParams, opts...)
 	defer stream.Close()
 
 	for stream.Next() {
-		if err := events.Send(ctx, respPayload.readDelta()); err != nil {
+		var ev responses.ResponseStreamEventUnion
+		ev = stream.Current()
+
+		if responseID == "" && ev.Response.ID != "" {
```
dannykopping: Needs a comment.

intercept/responses/base.go
```diff
@@ -158,6 +160,79 @@ func (i *responsesInterceptionBase) handleUpstreamError(ctx context.Context, ups
...
+}
+
+func (i *responsesInterceptionBase) recordUserPrompt(ctx context.Context, responseID string) {
+	prompt, err := i.lastUserPrompt()
+	if err != nil {
+		i.logger.Warn(ctx, "failed to get last user prompt", slog.Error(err))
+		return
+	}
+
+	if prompt != "" && responseID != "" {
```
dannykopping: Nit: keep the happy path unindented.

dannykopping (commented):
Looking good.

I appreciate the refactorings, but they must be moved to a separate PR in the stack please. Let's try keep PRs focused. Stacks make this relatively cheap.

mtojek (approved):
Agree with Danny's comment about the alternative approach to ship these changes in a separate PR, but it isn't crucial for me. I'd like to see some tests around `blocking.go` and `streaming.go`, but I suspect these will be addressed in another part of the stack.

PR #111, opened 2026-01-13
Extend circuit breaker functionality to support per-model isolation

kacpersaw:
## Summary

Extends the circuit breaker implementation to isolate failures at the **model level**, not just per-endpoint. This prevents a failing model (e.g., `claude-sonnet-4-20250514` being rate-limited) from blocking requests to other models on the same endpoint.

### Problem

With endpoint-only circuit breakers, if one model starts failing (e.g., due to rate limiting), all requests to that endpoint get blocked - even for healthy models.

### Solution

Circuit breakers are now keyed by `endpoint:model` tuple instead of just `endpoint`. Each model gets independent failure tracking and circuit state.

### Changes

- **Circuit breaker key**: Changed from `endpoint` to `endpoint:model`
- **Metrics labels**: Added `model` label to all circuit breaker metrics
- **Architecture**: Replaced middleware pattern with `Execute()` method to access model after request parsing
- **Tests**: Added per-model isolation tests

### Example

```
/v1/messages:claude-sonnet-4-20250514  → OPEN (rate limited)
/v1/messages:claude-3-5-haiku-20241022 → CLOSED (healthy)
```

Requests to haiku continue working even when sonnet is blocked.

### Files Changed

| File | Change |
|------|--------|
| `circuitbreaker/circuitbreaker.go` | `Get(endpoint, model)`, `Execute()` method |
| `bridge.go` | Pass model to circuit breaker after interceptor creation |
| `metrics/metrics.go` | Added `model` label to CB metrics |
| `circuitbreaker/circuitbreaker_test.go` | Per-model isolation tests |
| `circuit_breaker_integration_test.go` | Integration test for model isolation |

### DIFF
--- a/circuitbreaker/circuitbreaker.go
+++ b/circuitbreaker/circuitbreaker.go
@@ -12,6 +12,10 @@ import (
 	"github.com/sony/gobreaker/v2"
 )
 
+// ErrCircuitOpen is returned by Execute when the circuit breaker is open
+// and the request was rejected without calling the handler.
+var ErrCircuitOpen = errors.New("circuit breaker is open")
+
 // DefaultIsFailure returns true for standard HTTP status codes that typically
 // indicate upstream overload.
 func DefaultIsFailure(statusCode int) bool {
@@ -130,7 +134,8 @@ func (w *statusCapturingWriter) Unwrap() http.ResponseWriter {
 }
 
 // Execute runs the given handler function within circuit breaker protection.
-// If the circuit is open, the request is rejected with a 503 response and metrics are recorded.
+// If the circuit is open, the request is rejected with a 503 response, metrics are recorded,
+// and ErrCircuitOpen is returned.
 // Otherwise, it returns the handler's error (or nil on success).
 // The handler receives a wrapped ResponseWriter that captures the status code.
 // If the receiver is nil (no circuit breaker configured), the handler is called directly.
@@ -161,7 +166,7 @@ func (p *ProviderCircuitBreakers) Execute(endpoint, model string, w http.Respons
 		w.Header().Set("Retry-After", fmt.Sprintf("%d", int64(p.config.Timeout.Seconds())))
 		w.WriteHeader(http.StatusServiceUnavailable)
 		_, _ = w.Write(p.openErrorResponse())
-		return nil
+		return ErrCircuitOpen
 	}
 
 	return handlerErr
--- a/circuitbreaker/circuitbreaker_test.go
+++ b/circuitbreaker/circuitbreaker_test.go
@@ -1,6 +1,7 @@
 package circuitbreaker
 
 import (
+	"errors"
 	"net/http"
 	"net/http/httptest"
 	"sync/atomic"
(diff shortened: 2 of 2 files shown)


circuit_breaker_integration_test.go
```diff
@@ -535,9 +545,140 @@ func TestCircuitBreaker_HalfOpenMaxRequests(t *testing.T) {
...
+	require.NoError(t, err)
+
+	mockSrv := httptest.NewUnstartedServer(bridge)
+	t.Cleanup(mockSrv.Close)
+	mockSrv.Config.BaseContext = func(_ net.Listener) context.Context {
+		return aibridge.AsActor(ctx, "test-user-id", nil)
+	}
+	mockSrv.Start()
+
+	makeRequest := func(model string) *http.Response {
```
pawbana: `createAnthropicMessagesReq` and `createOpenAIChatCompletionsReq` functions could be used for creating requests. Please use them to limit code duplication, also in previously written tests.

bridge.go
```diff
@@ -202,18 +197,27 @@ func newInterceptionProcessor(p provider.Provider, rec recorder.Recorder, mcpPro
...
 
-		if err := interceptor.ProcessRequest(w, r); err != nil {
+		// Process request with circuit breaker protection if configured
+		switch err := cbs.Execute(route, interceptor.Model(), w, func(rw http.ResponseWriter) error {
+			return interceptor.ProcessRequest(rw, r)
+		}); {
+		case errors.Is(err, circuitbreaker.ErrCircuitOpen):
+			if m != nil {
+				m.CircuitBreakerRejects.WithLabelValues(p.Name(), route, interceptor.Model()).Inc()
+			}
```
pawbana: This check could be moved into `Execute`, laving if else statement as it was. It adds metrics parameter to `Execute` signature but I think it makes sense to update metrics in there.

circuitbreaker/circuitbreaker.go
```diff
@@ -153,11 +154,14 @@ func (p *ProviderCircuitBreakers) Execute(endpoint, model string, w http.Respons
...
 	if errors.Is(err, gobreaker.ErrOpenState) || errors.Is(err, gobreaker.ErrTooManyRequests) {
+		if p.metrics != nil {
+			p.metrics.CircuitBreakerRejects.WithLabelValues(p.provider, endpoint, model).Inc()
+		}
 		w.Header().Set("Content-Type", "application/json")
 		w.Header().Set("Retry-After", fmt.Sprintf("%d", int64(p.config.Timeout.Seconds())))
 		w.WriteHeader(http.StatusServiceUnavailable)
 		_, _ = w.Write(p.openErrorResponse())
-		return ErrCircuitOpen
+		return nil
```
pawbana: I think `ErrCircuitOpen` error should be returned. I would interpret circuit breaker rejecting request the same as too eg. many request response from upstream before the change which I believe causes `ProcessRequest` to return error (at least I think it happens in blocking case, may not be happening in streaming case).

Issue #126, opened 2026-01-16
Support customizing Bedrock endpoint url & custom headers


dannykopping:
A user reported needing to customize the Bedrock endpoint to relay requests through an internal proxy which ultimately relays requests to Bedrock.

They are currently using Claude Code to interact with this proxy by using the following env vars:

- `CLAUDE_CODE_USE_BEDROCK=1`
- `CLAUDE_CODE_SKIP_BEDROCK_AUTH=1`
- `ANTHROPIC_BEDROCK_BASE_URL=corp.proxy`

---

Additionally they need to be able to set static HTTP headers to go along with each request. See https://github.com/coder/aibridge/issues/135.

PR #131, opened 2026-01-20
feat: allow bedrock base url definition

dannykopping:
Partially addresses https://github.com/coder/aibridge/issues/126


ssncferreira (approved):
LGTM 👍 just a question, what happens if the user configures both a `Region` and a `BaseURL` for Bedrock? My guess is that the BaseURL takes precedence (since it's appended last to `opts`. Maybe, a test validating the precedence behavior would be helpful, wdyt?

PR #134, opened 2026-01-20
fix: stop forwarding empty response when no response was received

pawbana:
Fixes edge case in responses blocking interceptor.
When no response was received from upstream, first custom error was sent then empty / not initialized copy of the response was forwarded.
Fix is to not sent empty copies of responses.


dannykopping (commented):
Can you please add a test for this scenario?

Issue #20, opened 2025-10-03
Identify client by User-Agent or other means


dannykopping:
We should track which client applications are being used to interface with aibridge.

Claude Code, for example, helpfully sets a custom UA: `claude-cli/2.0.1 (external, cli)`.
Cursor, last time I checked, just left it as [`axios`](https://github.com/axios/axios); we may need to use other methods/heuristics to identify this client.

We'll need to enhance the recording API to pass this information along on the `InterceptionRecord`.

dannykopping:
Nice 👍 things have improved since I last looked!
Thanks for doing the legwork
This looks like a reasonable place to start

matifali:
These are the only ones I tried. We can look into more when we implement.

dannykopping:
@matifali @pawbana WDYT?

matifali:
> 3 can be a follow-up since it's not part of this issue's scope, but I'm not really sure we need these. It will ~10x the cardinality on the metrics it's added to for little value.

What do you think is the recommended way for users wanting to see this information in their monitoring system? We can file a separate issue, but we should provide them with a way to do so. I understand that a Prometheus metric may not be the best choice here, so I am asking for alternate suggestions.

PR #158, opened 2026-01-29
feat: record user agent and best effort client

pawbana:
Adds raw user agent value and heuristically guessed client name to interception recording.

Fixes: https://github.com/coder/aibridge/issues/31

### DIFF
--- a/bridge.go
+++ b/bridge.go
@@ -24,6 +24,21 @@ import (
 	"go.opentelemetry.io/otel/trace"
 )
 
+const (
+	// The duration after which an async recording will be aborted.
+	recordingTimeout = time.Second * 5
+
+	ClientClaude     = "Claude Code"
+	ClientCodex      = "Codex"
+	ClientZed        = "Zed"
+	ClientCopilotVSC = "GitHub Copilot (VS Code)"
+	ClientCopilotCLI = "GitHub Copilot (CLI)"
+	ClientKilo       = "Kilo Code"
+	ClientRoo        = "Roo Code"
+	ClientCursor     = "Cursor"
+	ClientUnknown    = "Unknown"
+)
+
 // RequestBridge is an [http.Handler] which is capable of masquerading as AI providers' APIs;
 // specifically, OpenAI's & Anthropic's at present.
 // RequestBridge intercepts requests to - and responses from - these upstream services to provide
@@ -51,9 +66,6 @@ type RequestBridge struct {
 }
 
 var _ http.Handler = &RequestBridge{}
-
-// The duration after which an async recording will be aborted.
-const recordingTimeout = time.Second * 5
 
 // NewRequestBridge creates a new *[RequestBridge] and registers the HTTP routes defined by the given providers.
 // Any routes which are requested but not registered will be reverse-proxied to the upstream service.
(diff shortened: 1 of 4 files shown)


pawbana:
* **#158** <a href="https://app.graphite.com/github/pr/coder/aibridge/158?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/aibridge/158?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

bridge.go
```diff
@@ -299,3 +301,33 @@ func mergeContexts(base, other context.Context) context.Context {
...
+// guessClient attempts to guess the client application from the request headers.
+// Not all clients set proper user agent headers, so this is a best-effort approach.
+// Based on https://github.com/coder/aibridge/issues/20#issuecomment-3769444101.
+func guessClient(r *http.Request) string {
+	userAgent := strings.ToLower(r.UserAgent())
+	originator := r.Header.Get("originator")
+
+	switch {
+	case strings.Contains(userAgent, "claude"):
+		return "Claude Code"
```
dannykopping: Let's make these constants so we can reuse them in the tests.

matifali:
> What more would you need to see added for this to be considered "complete system of identification" @matifali?

See my commnet on issue: https://github.com/coder/aibridge/issues/20#issuecomment-3848104603

bridge.go
```diff
@@ -299,3 +301,33 @@ func mergeContexts(base, other context.Context) context.Context {
...
+	case strings.Contains(userAgent, "zed"):
+		return "Zed"
+	case strings.Contains(userAgent, "copilot"):
+		return "GitHubCopilot"
+	case strings.Contains(userAgent, "kilo-code") || strings.Contains(originator, "kilo-code"):
+		return "Kilo Code"
+	case strings.Contains(userAgent, "roo-code") || strings.Contains(originator, "roo-code"):
+		return "Roo Code"
+	default:
+		for key := range r.Header {
```
ssncferreira: Wouldn't it be possible to use the exact header that might have `x-cursor`? This will be O(n) on every request where none of the earlier cases match 😕

ssncferreira (approved):
LGTM 👍

recorder/types.go
```diff
@@ -26,10 +26,14 @@ type ToolArgs any
...
-	Metadata                     Metadata
-	StartedAt                    time.Time
+	Client      string
+	ID          string
+	InitiatorID string
+	Metadata    Metadata
+	Model       string
+	Provider    string
+	StartedAt   time.Time
+	UserAgent   string
```
ssncferreira: Since we already have `Client` do we need to store this? Wouldn't it be possible to add it to `Metadata`?

bridge.go
```diff
@@ -23,6 +23,21 @@ import (
...
+	ClientClaude     = "Claude Code"
+	ClientCodex      = "Codex"
+	ClientZed        = "Zed"
+	ClientCopilotVSC = "GitHub Copilot (VS Code)"
+	ClientCopilotCLI = "GitHub Copilot (CLI)"
+	ClientKilo       = "Kilo Code"
+	ClientRoo        = "Roo Code"
+	ClientCursor     = "Cursor"
+	ClientUnknown    = "Unknown"
+)
```
matifali: Can we add `cline` and `JetBrains` to the list?

There is some investigation **here**

bridge.go
```diff
@@ -299,3 +313,33 @@ func mergeContexts(base, other context.Context) context.Context {
...
+	case strings.HasPrefix(userAgent, "zed/"):
+		return ClientZed
+	case strings.HasPrefix(userAgent, "githubcopilotchat/"):
+		return ClientCopilotVSC
+	case strings.HasPrefix(userAgent, "copilot/"):
+		return ClientCopilotCLI
+	case strings.HasPrefix(userAgent, "kilo-code/") || originator == "kilo-code":
+		return ClientKilo
+	case strings.HasPrefix(userAgent, "roo-code/") || originator == "roo-code":
+		return ClientRoo
```
matifali: When Provider is Anthropic Both kilo code and Roo code have a different agent
. Please see: https://github.com/coder/aibridge/issues/20#issuecomment-3769444101

recorder/types.go
pawbana: In general I'm open to adding user agent as a field to metadata and I agree that it would simplify a bit code in coder PRs but since we have separate repos I'd like to maintain "neutrality" and rather not think about coder implementation details in AI Bridge PRs. Maybe I'm a bit too strict with this approach.

I think this approach makes it more flexible. If AI Bridge user wants to store user-agent in metadata it is possible (how later PRs are doing) if they rather store it differently it is also possible without deleting things from metadata. With separate fields there is clear boundary `Metadata` == what was set in `AsActor.Metadata`. If user would like to have it in metadata it could be added sooner in `AsActor`. Maybe this would be better approach (although then each user would have to extract user-agent himself)?.
Now that I look at it, it seems a bit unnecessary that metadata is simply going back and forth but I think there will be some use for it later.

bridge.go
pawbana: Maybe I'm too defensive here but I think we should set client only when we have very high confidence the client is correct, so when user agent (or some other header) uniquely identifies application.

My assumption is that it would be better to have as little false positives as possible, even with the cost of not recording JetBrains 90% of the time correctly.  For investigating there is fallback with raw user agent being added to metadata if needed.

bridge.go
```diff
@@ -318,3 +332,33 @@ func mergeContexts(base, other context.Context) context.Context {
...
+	case strings.HasPrefix(userAgent, "copilot"):
+		return ClientCursor
```
pymq: @pawbana Hi, are these two lines a typo? (by prefix `copilot` return `ClientCursor`)

Especially because there is already a check for the `copilot/` prefix above

(I was just randomly passing by)

PR #164, opened 2026-02-04
fix: preserve the stream property for chat/completions calls

SasSwart:
In the `openai-go` SDK, a conflict between a [recent optimization](https://github.com/SasSwart/openai-go/pull/1) and the way in which options are set resulted in an issue whereby request bodies for `chat/completions` calls were being discarded.

This PR fixes the issue by taking full control of the request body handling. We now use the `WithRequestBody` and `WithJSONSet` options ourselves to ensure that the request payload is sent exactly as we need it.

Integration tests have been added to guard against future regression.

Related to [internal Slack thread](https://codercom.slack.com/archives/C096PFVBZKN/p1770195508426909).

### DIFF
--- a/intercept/chatcompletions/streaming.go
+++ b/intercept/chatcompletions/streaming.go
@@ -125,6 +125,9 @@ func (i *StreamingInterception) ProcessRequest(w http.ResponseWriter, r *http.Re
 			opts = append(opts, intercept.ActorHeadersAsOpenAIOpts(actor)...)
 		}
 
+		// We take control of request body here and pass it to the SDK as a raw byte slice.
+		// This is because the SDK's serialization applies hidden request options that result in
+		// unexpected, breaking behaviour. See https://github.com/coder/aibridge/pull/164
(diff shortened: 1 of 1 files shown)


intercept/chatcompletions/streaming.go
```diff
@@ -125,6 +125,13 @@ func (i *StreamingInterception) ProcessRequest(w http.ResponseWriter, r *http.Re
 			opts = append(opts, intercept.ActorHeadersAsOpenAIOpts(actor)...)
 		}
 
+		body, err := json.Marshal(i.req.ChatCompletionNewParams)
```
ssncferreira: nit: would be nice to have a small comment explaining why this is needed

SasSwart: Done in [a787d86](https://github.com/coder/aibridge/pull/164/commits/a787d862e9bbe8a40ee973c9ccda4383799ecc35)

Issue #166, opened 2026-02-05
Record session ID


dannykopping:
See [https://www.notion.so/coderhq/AI-Bridge-Session-Tracking-2f7d579be59280799a61e04d34005a5a?source=copy_link](<https://www.notion.so/coderhq/AI-Bridge-Session-Tracking-2f7d579be59280799a61e04d34005a5a?source=copy_link>)

Claude Code and Codex send through "session" IDs which indicate that an interception belongs to a conversation (series of related requests/responses) within that tool.

Acceptance criteria:

- Session ID is determined from all interceptor implementations (where available)
  - [X] Messages
  - [X] Chat Completions
  - [X] Responses
- [X] Recorder types updated
- [X] Tests added

PR #186, opened 2026-02-19
feat: add upstream request/response logging for passthrough routes

pawbana:
Added request/response dump functionality for passthrough routes.

- Added  `RoundTripperMiddleware` in the `apidump` package.
- Updated the Provider interface to expose the `APIDumpDir` method

Drive by change
- Improved error handling in the dumper code

### DIFF
--- a/intercept/apidump/apidump.go
+++ b/intercept/apidump/apidump.go
@@ -32,10 +32,9 @@ type MiddlewareNext = func(*http.Request) (*http.Response, error)
 // Middleware is an HTTP middleware function compatible with SDK WithMiddleware options.
 type Middleware = func(*http.Request, MiddlewareNext) (*http.Response, error)
 
-// NewMiddleware returns a middleware function that dumps requests and responses to files.
-// Files are written to the path returned by DumpPath.
+// NewBridgeMiddleware returns a middleware function that dumps requests and responses to files.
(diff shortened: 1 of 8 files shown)


intercept/apidump/apidump.go
```diff
@@ -175,16 +188,72 @@ func (d *dumper) writeRedactedHeaders(w io.Writer, headers http.Header, sensitiv
...
+}
+
+// passthroughDumpPath returns the base file path (without suffix) for a passthrough dump.
+// A random UUID is generated for the filename. "passthrough" is used as the directory name
+// in place of the model.
+func passthroughDumpPath(baseDir string, provider string, clk quartz.Clock) string {
+	return filepath.Join(baseDir, provider, "passthrough", fmt.Sprintf("%d-%s", clk.Now().UTC().UnixMilli(), uuid.New()))
+}
+
+// NewRoundTripperMiddleware returns http.RoundTripper that dumps requests and responses to files.
```
dannykopping: This needs a more descriptive name please.

dannykopping:
> > I'm thinking we should perhaps hold off on this change until #166 lands. If we have a session ID then we can correlate all intercepted and passthrough routes.
> 
> Will session ID be available in each request? Or only in requests from clients that set it?

Only if clients set it

Issue #192, opened 2026-02-26
Forward client request headers to upstream providers


ssncferreira:
## Problem

AI Bridge drops client request headers when making upstream API calls. The messages interceptor constructs a new HTTP request via the provider SDK, which sets its own headers (SDK user-agent, API key, version, etc.) and does not include the original client headers. While we support setting individual headers via `option.WithHeader`, this approach is cumbersome as it requires code changes every time a new header is required.

For example, Claude Code sends `Anthropic-Beta: context-management` header alongside a `context_management` field in the request body. AI Bridge preserves the body but drops the header, causing Anthropic to reject the request with a 400. The current workaround is `CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS=1`, which prevents Claude Code from sending the beta field entirely. Related to internal Slack thread: https://codercom.slack.com/archives/C096PFVBZKN/p1772115088945279

## Proposed solution

Bridged routes should forward client headers to the upstream provider, similar to what passthrough routes already do via `req.Header = r.Header.Clone()`.

PR #205, opened 2026-03-06
fix: forward Anthropic-Beta header to Anthropic provider

ssncferreira:
## Description

AI Bridge drops client request headers when making upstream API calls through bridge routes. This causes failures when the client sends headers that correspond to features used in the request body. For example, Claude Code sends `Anthropic-Beta: context-management` alongside a `context_management` field in the body. AI Bridge preserves the body but drops the header, causing Anthropic to reject the request.

Rather than forwarding all client headers (addressed in https://github.com/coder/aibridge/pull/195), this PR takes a targeted approach by forwarding only known required headers for Anthropic, matching the same pattern used by the Copilot provider with `ExtraHeaders`.

Related to https://github.com/coder/aibridge/issues/192

**Note:** This is an intermediary fix. The full solution, forwarding all client headers to upstream providers, is tracked in https://github.com/coder/aibridge/pull/195. Related to internal Slack thread: https://codercom.slack.com/archives/C096PFVBZKN/p1772792545309049

### DIFF
--- a/provider/anthropic_test.go
+++ b/provider/anthropic_test.go
@@ -1,11 +1,170 @@
 package provider
 
 import (
+	"bytes"
 	"net/http"
+	"net/http/httptest"
 	"testing"
 
+	"cdr.dev/slog/v3"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/coder/aibridge/config"
+	"github.com/coder/aibridge/internal/testutil"
 )
+
+func TestAnthropic_CreateInterceptor(t *testing.T) {
+	t.Parallel()
+
+	provider := NewAnthropic(config.Anthropic{Key: "test-key"}, nil)
+
+	t.Run("Messages_NonStreamingRequest_BlockingInterceptor", func(t *testing.T) {
+		t.Parallel()
+
+		body := `{"model": "claude-opus-4-5", "max_tokens": 1024, "messages": [{"role": "user", "content": "hello"}], "stream": false}`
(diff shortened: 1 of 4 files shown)


provider/anthropic_test.go
```diff
@@ -1,12 +1,171 @@
...
+		w := httptest.NewRecorder()
+
+		interceptor, err := provider.CreateInterceptor(w, req, testTracer)
+
+		require.Error(t, err)
+		require.Nil(t, interceptor)
+		assert.Contains(t, err.Error(), "unmarshal request body")
+	})
+
+	t.Run("Messages_ForwardsAnthropicBetaHeaderToUpstream", func(t *testing.T) {
```
dannykopping: Nit: AFAICS this is the only relevant sub-test.
Adding the others is a nice improvement, though.

provider/anthropic_test.go
```diff
@@ -1,12 +1,171 @@
...
 
+	"cdr.dev/slog/v3"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/coder/aibridge/config"
+	"github.com/coder/aibridge/internal/testutil"
 )
 
+func TestAnthropic_CreateInterceptor(t *testing.T) {
```
pawbana: Those test look quite copy-pasty, eg. body in each test case looks the same or verification that interceptor was created without error.
Would it be possible to shorten them with table test approach?

pawbana (approved):

PR #207, opened 2026-03-06
fix: sync example module dependencies with root

ssncferreira:
The `example/` module dependencies were out of sync with the root module. 

Note: GoLand auto-indexes nested modules, which triggered `go mod tidy` and caused unexpected diffs leaking into unrelated PRs.

### DIFF
--- a/example/go.mod
+++ b/example/go.mod
@@ -3,7 +3,7 @@ module github.com/coder/aibridge/example
 go 1.25.6
 
 require (
-	cdr.dev/slog/v3 v3.0.0-rc1
+	cdr.dev/slog/v3 v3.0.0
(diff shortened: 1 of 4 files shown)


example/go.mod
```diff
@@ -1,15 +1,13 @@
 module github.com/coder/aibridge/example
 
-go 1.24.6
-
-toolchain go1.24.10
+go 1.25.6
 
 require (
-	cdr.dev/slog v1.6.2-0.20250703074222-9df5e0a6c145
+	cdr.dev/slog/v3 v3.0.0-rc1
```
dannykopping: Nit: v3.0.0 is out, we don't need to use the RC

PR #214, opened 2026-03-12
feat: forward client request headers to upstream providers in bridge routes

ssncferreira:
## Description

AI Bridge acts as a reverse proxy between clients and upstream AI providers. However, bridge routes were dropping client request headers as the interceptors construct new HTTP requests via provider SDKs, which set their own headers and discard the originals.

This PR changes bridge routes to forward client request headers to upstream providers, so the upstream sees the same headers the client sent, except headers related to auth, transport, or managed by aibridge.

## Changes

* Add `intercept/client_headers.go` with `SanitizeClientHeaders` and `BuildUpstreamHeaders`. Client headers are sanitized by stripping hop-by-hop ([RFC 2616](https://www.rfc-editor.org/rfc/rfc2616#section-13.5.1)), transport (`Host`, `Accept-Encoding`, `Content-Length`), and auth (`Authorization`, `X-Api-Key`) headers.
* Add SDK middleware in all three interceptor base types that replaces the SDK-built request headers with the sanitized client headers. Two categories of headers are preserved from the SDK-built request:
  * Auth headers: the SDK sets the correct provider credentials (API key or per-user token for Copilot), and these must not be overwritten by the client's auth header.
  * Actor headers (`X-AI-Bridge-Actor-*`): injected by aibridge as per-request SDK options to identify the requesting user to the upstream.
* The client-header middleware is appended before the API dump middleware, so that API dump captures the final outgoing headers as they are sent to the upstream.

This is an intermediate step toward removing the SDK from the HTTP transport path. By forwarding client headers directly, aibridge is no longer dependent on SDK-managed headers, making future SDK removal simpler.

## Follow-up

* Proxy headers: Passthrough routes set standard proxy headers (`X-Forwarded-For`, `X-Forwarded-Host`, `X-Forwarded-Proto`), while bridge routes do not. Since AI Bridge behaves as a reverse proxy, this handling should be consistent across all requests.
* Inject actor headers directly in the client-header middleware instead of using per-request SDK options.
* Remove `ExtraHeaders` mechanism for Anthropic and Copilot: now redundant since all client headers are forwarded.

## Tests

Tested manually with:
* ✅ Claude Code 2.1.74
* ✅ codex-cli 0.114.0
* ✅ Copilot VS Code 0.38.2 (via AI Bridge Proxy)

Closes: https://github.com/coder/aibridge/issues/192

### DIFF
--- a/provider/openai_test.go
+++ b/provider/openai_test.go
@@ -162,89 +162,68 @@ func generateResponsesPayload(payloadSize int, inputCount int, stream bool) []by
 func TestOpenAI_CreateInterceptor(t *testing.T) {
 	t.Parallel()
 
-	t.Run("ChatCompletions_ClientHeaders", func(t *testing.T) {
-		t.Parallel()
+	tests := []struct {
+		name         string
+		route        string
+		requestBody  string
+		responseBody string
+	}{
+		{
+			name:         "ChatCompletions_ClientHeaders",
+			route:        routeChatCompletions,
+			requestBody:  `{"model": "gpt-4", "messages": [{"role": "user", "content": "hello"}], "stream": false}`,
+			responseBody: chatCompletionResponse,
+		},
+		{
+			name:         "Responses_ClientHeaders",
+			route:        routeResponses,
+			requestBody:  `{"model": "gpt-5", "input": "hello", "stream": false}`,
+			responseBody: responsesAPIResponse,
+		},
+	}
 
-		var receivedHeaders http.Header
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
 
-		mockUpstream := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			receivedHeaders = r.Header.Clone()
-			w.Header().Set("Content-Type", "application/json")
-			w.WriteHeader(http.StatusOK)
-			_, err := w.Write([]byte(chatCompletionResponse))
+			var receivedHeaders http.Header
+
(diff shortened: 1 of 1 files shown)


intercept/client_headers.go
```diff
@@ -0,0 +1,70 @@
...
+// nonForwardedHeaders are transport-level headers managed by aibridge or
+// Go's HTTP transport that must not be forwarded to the upstream provider.
+var nonForwardedHeaders = []string{
+	"Host",
+	"Accept-Encoding",
+	"Content-Length",
+}
+
+// authHeaders are headers that carry authentication credentials from the
+// client. These are stripped because the SDK re-injects the correct
```
dannykopping: "the SDK re-injects" can be reworded to be more explicit.  
Which SDK? Where does it do that?  
  
(i know the answers but future readers may not)

provider/openai_test.go
```diff
@@ -150,6 +154,92 @@ func generateResponsesPayload(payloadSize int, inputCount int, stream bool) []by
...
+
+	t.Run("ChatCompletions_ClientHeaders", func(t *testing.T) {
+		t.Parallel()
+
+		var receivedHeaders http.Header
+
+		mockUpstream := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			receivedHeaders = r.Header.Clone()
+			w.Header().Set("Content-Type", "application/json")
+			w.WriteHeader(http.StatusOK)
```
dannykopping: Nit: can we use a fixture instead? It'll make the tests a bit more concise.

PR #216, opened 2026-03-12
feat: aibridge BYOK for Anthropic

evgeniy-scherbina:
### Changes

  **aibridge:**

  - `config/config.go` — Added `BYOKBearerToken` field to `Anthropic` config for OAuth-based BYOK (Claude Max/Pro).
  - `provider/anthropic.go` — `CreateInterceptor` detects user LLM credentials from surviving request headers and sets them on the config copy. `InjectAuthHeader` skips centralized key injection when user credentials are already present (passthrough BYOK).
  - `intercept/messages/base.go` — `newMessagesService` uses `option.WithAuthToken()` (Authorization: Bearer) when `BYOKBearerToken` is set, otherwise `option.WithAPIKey()` (X-Api-Key).
  
  ### How the flow works end-to-end

  **Centralized:** Client sends `Authorization: Bearer <coder-token>` → http.go extracts token, strips all auth headers → providers get no user credentials → inject centralized key.

  **BYOK (Claude Max/Pro):** Client sends `Authorization: Bearer <oauth-token>` \+ `X-Coder-AI-Governance-BYOK-Token: <coder-token>` → http.go extracts Coder token from BYOK header, strips only BYOK header → `Authorization: Bearer <oauth-token>` survives → `CreateInterceptor` picks it up as `BYOKBearerToken` → SDK sends it via `WithAuthToken()`. Passthrough: `InjectAuthHeader` sees existing `Authorization`, skips injection.

  **BYOK (personal API key):** Client sends `X-Api-Key: <api-key>` + `X-Coder-AI-Governance-BYOK-Token: <coder-token>` → same flow but `CreateInterceptor` picks up `X-Api-Key` as `cfg.Key` → SDK sends it via `WithAPIKey()`. Passthrough: `InjectAuthHeader` sees existing `X-Api-Key`, skips injection.
  
### Scope

  This PR covers BYOK for the Anthropic provider only. OpenAI provider BYOK support and the key-type observability requirement (Requirement 2 in the RFC) are not included and will be addressed in follow-up PRs.


provider/anthropic.go
```diff
@@ -110,11 +110,31 @@ func (p *Anthropic) CreateInterceptor(w http.ResponseWriter, r *http.Request, tr
 		cfg := p.cfg
 		cfg.ExtraHeaders = extractAnthropicHeaders(r)
 
+		// In centralized mode, http.go strips Authorization and X-Api-Key
+		// (they carried the Coder token), so neither header is present
+		// here and cfg keeps the centralized key.
```
ssncferreira: I'm confused by this comment, is this a reference to the coder repo? If yes, I would suggest not using `http.go` and making it clear that the upstream caller needs to strip `Authorization` and `X-API-Key` headers.

IIUC, at this point, the headers are as follows:
1) Centralized: no Authorization header or X-API-Key
2) BYOK (oauth): Authorization header includes user's oauth token, no X-API-key
3) BYOK Api key: no Authorization header, X-API-Key is user's API key

provider/anthropic.go
```diff
@@ -110,11 +110,31 @@ func (p *Anthropic) CreateInterceptor(w http.ResponseWriter, r *http.Request, tr
...
+		//   - X-Api-Key: <api-key> → personal API key. Overwrite the
+		//     centralized key with the user's key.
+		authHeaderName := p.AuthHeader()
+		if bearer := r.Header.Get("Authorization"); bearer != "" {
+			cfg.BYOKBearerToken = strings.TrimPrefix(bearer, "Bearer ")
+			cfg.Key = ""
+			authHeaderName = "Authorization"
+		} else if apiKey := r.Header.Get("X-Api-Key"); apiKey != "" {
+			cfg.Key = apiKey
+		}
```
ssncferreira: From this point on, we no longer know whether this interception is using a centralized (global) key or a BYOK (user's personal) API key, right? This could be useful to store and to show in the logs (the same for BYOK oauth token). For example, if Anthropic returns a 401, we wouldn't know if the failing key is the global key (affecting everyone) or a single user's personal key.

Additionally, this is probably out of scope for this PR, but it might make sense to store this information in the interception so we can later surface it in the UI, wdyt?

utils/mask.go
```diff
@@ -0,0 +1,10 @@
+package utils
+
+// MaskSecret returns the first 4 and last 4 characters of s
+// separated by "...", or the full string if 8 characters or fewer.
+func MaskSecret(s string) string {
```
ssncferreira, line 8: Is this a good idea? I think logging the auth mode ("centralized", "byok_bearer", "byok_apikey") rather than a hint of the secret might be cleaner 👀
Additionally, if we need to correlate a failure to a specific user, I believe we already log this in some cases.

provider/anthropic.go
evgeniy-scherbina: @ssncferreira  I’ve updated the comment - feel free to suggest any improvements.

Your understanding of the headers is correct.

provider/anthropic.go
evgeniy-scherbina: Good comment - I’ll address it in the next PR, if that’s okay.

> we no longer know whether this interception is using a centralized (global) key or a BYOK (user's personal) API key

Yes, that's correct.

utils/mask.go
evgeniy-scherbina: I will definitely log and store `auth mode` in the next PR. But as we discussed we can have multiple personal api keys and multiple centralized keys, and we want to be able to identify what key was used.

config/config.go
```diff
@@ -15,6 +15,10 @@ type Anthropic struct {
 	CircuitBreaker   *CircuitBreaker
 	SendActorHeaders bool
 	ExtraHeaders     map[string]string
+	// BYOKBearerToken is set in BYOK mode when the user authenticates
+	// with an OAuth token (e.g. Claude Max/Pro subscription). When set,
```
dannykopping: Nit: I don't think we should specifically call out any specific provider here; this is provider-agnostic.
Please replace this throughout.

config/config.go
```diff
@@ -15,6 +15,10 @@ type Anthropic struct {
...
+	// the SDK uses Authorization: Bearer instead of X-Api-Key.
```
dannykopping: Which SDK?

intercept/messages/base.go
```diff
@@ -205,7 +206,19 @@ func (i *interceptionBase) isSmallFastModel() bool {
 }
 
 func (i *interceptionBase) newMessagesService(ctx context.Context, opts ...option.RequestOption) (anthropic.MessageService, error) {
-	opts = append(opts, option.WithAPIKey(i.cfg.Key))
+	// BYOK with OAuth token (Claude Max/Pro) uses Authorization: Bearer.
+	// Otherwise use X-Api-Key (centralized or BYOK with personal API key).
+	if i.cfg.BYOKBearerToken != "" {
+		i.logger.Debug(ctx, "using byok oauth bearer auth",
+			slog.F("bearer_hint", utils.MaskSecret(i.cfg.BYOKBearerToken)),
```
dannykopping, line 213: This will not be enough information for an operator to use in any way.
I actually don't think logging at this point is very useful, TBH; we should instead when an interception is created. We at least have to log the interception ID, client used, etc.

provider/anthropic.go
```diff
@@ -110,11 +110,35 @@ func (p *Anthropic) CreateInterceptor(w http.ResponseWriter, r *http.Request, tr
 		cfg := p.cfg
 		cfg.ExtraHeaders = extractAnthropicHeaders(r)
 
+		// coder/aibridged strips all headers that may carry the Coder
```
dannykopping: Nit: I'd rather we didn't document external systems' behaviour here since it can change.

provider/anthropic.go
```diff
@@ -110,11 +110,35 @@ func (p *Anthropic) CreateInterceptor(w http.ResponseWriter, r *http.Request, tr
...
+		// session token before passing the request here, so this code
+		// sees only legitimate LLM credentials.
+		//
+		// In centralized mode neither Authorization nor X-Api-Key is
+		// present, so cfg keeps the centralized key unchanged.
+		//
+		// In BYOK mode the user's LLM credentials survive intact.
+		// Either Authorization or X-Api-Key will be present, but not
+		// both. If Authorization is present it means the user has a
```
dannykopping: What if they are both set?

provider/anthropic_test.go
```diff
@@ -105,11 +104,77 @@ func TestAnthropic_CreateInterceptor(t *testing.T) {
...
+		wantXApiKey       string
+		wantAuthorization string
+	}{
+		{
+			name:              "Messages_BYOK_BearerToken",
+			setHeaders:        map[string]string{"Authorization": "Bearer user-oauth-token"},
+			wantAuthorization: "Bearer user-oauth-token",
+		},
+		{
+			name:        "Messages_BYOK_APIKey",
```
dannykopping, line 143: We need a test here to define expected behaviour when both `Authorization` and `X-Api-Key` are set.

provider/anthropic_test.go
```diff
@@ -124,6 +189,52 @@ func TestAnthropic_CreateInterceptor(t *testing.T) {
...
+		},
+		{
+			name:          "pre-existing X-Api-Key is not overwritten",
+			presetHeaders: map[string]string{"X-Api-Key": "user-api-key"},
+			wantXApiKey:   "user-api-key",
+		},
+		{
+			name:              "pre-existing Authorization prevents centralized key injection",
+			presetHeaders:     map[string]string{"Authorization": "Bearer user-oauth-token"},
+			wantXApiKey:       "",
```
dannykopping: Nit: match pattern in above test of using zero value for uninvolved header.

```suggestion
```

provider/anthropic_test.go
```diff
@@ -124,6 +189,52 @@ func TestAnthropic_CreateInterceptor(t *testing.T) {
...
+		wantXApiKey       string
+		wantAuthorization string
+	}{
+		{
+			name:          "no pre-existing auth headers injects centralized key",
+			presetHeaders: map[string]string{},
+			wantXApiKey:   "centralized-key",
+		},
+		{
+			name:          "pre-existing X-Api-Key is not overwritten",
```
dannykopping: Please add a test for both headers here.

provider/anthropic_test.go
```diff
@@ -124,6 +189,52 @@ func TestAnthropic_CreateInterceptor(t *testing.T) {
...
+			presetHeaders: map[string]string{"X-Api-Key": "user-api-key"},
+			wantXApiKey:   "user-api-key",
+		},
+		{
+			name:              "pre-existing Authorization prevents centralized key injection",
```
dannykopping: Nit: the phrasing "is not overwritten" and "prevents centralized key injection" suggests there are two different behaviours, but there are not; it's the same behaviour.

utils/mask.go
```diff
@@ -0,0 +1,10 @@
...
+	if len(s) <= 8 {
+		return s
+	}
+	return s[:4] + "..." + s[len(s)-4:]
```
dannykopping: ‼️ If the secret is 8 chars or less it'll log the whole secret.

utils/mask_test.go
```diff
@@ -0,0 +1,30 @@
...
+func TestMaskSecret(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		input    string
+		expected string
+	}{
+		{"", ""},
+		{"short", "short"},
+		{"exactly8", "exactly8"},
```
dannykopping: ‼️ Why is it expected that we log the secret instead of masking?

dannykopping (changes requested):
Directionally looking fine but has a couple design issues.

Please improve the logging to be more diagnostically useful to operators and ensure we _never_ log the supplied BYOK secret.

passthrough.go
```diff
@@ -96,6 +97,16 @@ func newPassthroughRouter(provider provider.Provider, logger slog.Logger, m *met
 
 				// Inject provider auth.
 				provider.InjectAuthHeader(&req.Header)
+
+				if authz := req.Header.Get("Authorization"); authz != "" {
+					logger.Debug(ctx, "passthrough using oauth bearer auth",
+						slog.F("bearer_hint", utils.MaskSecret(authz)),
```
dannykopping: This will contain `Bearer ...` so it won't behave as expected.

dannykopping (commented):
This PR also needs to explicitly mention that it's for Anthropic _only_, please.

intercept/messages/base.go
evgeniy-scherbina: This is very helpful for debugging/investigation which key was used. It's the moment right before the SDK call where the auth decision materializes. Moving it to interception creation would mean logging the intent to use a key, not the actual key used. If something mutates cfg between creation and the SDK call, the earlier log would be misleading. Though it should not happen in practice.

<hr/>

The decision logic for selecting which key to use is provider-specific. I can move it into `Anthropic.CreateInterceptor` immediately after config.Anthropic is updated?

provider/anthropic.go
evgeniy-scherbina: The core issue is that the Authorization header is overloaded—it may contain either a Coder-Session-Token or an OAuth token intended for the LLM provider. As a result, when aibridge receives a request with an Authorization header, it cannot reliably determine whether the header should be forwarded or stripped.

To address this, we establish a clear convention: all Coder-Session-Token values are stripped at the coder/aibridge boundary. By the time the request reaches this layer, it should represent a standard LLM request.

This comment formalizes that expectation—any Coder-specific headers must be removed upstream, and if an Authorization header is present, it is assumed to contain a valid OAuth token for the LLM provider.

evgeniy-scherbina: @dannykopping I slightly updated comment:
```
// At this point the request contains only LLM provider headers.
// Any Coder-specific authentication has already been stripped.
```

If you want to remove all mentions of Coder (as an external system), I can do that. However, this code relies on Coder’s current behavior of stripping Coder-specific auth headers, so changes to that behavior could break it.

provider/anthropic.go
```diff
@@ -110,11 +110,34 @@ func (p *Anthropic) CreateInterceptor(w http.ResponseWriter, r *http.Request, tr
...
+		// Claude Max/Pro subscription and authenticated via OAuth;
+		// in this case set BYOKBearerToken so the SDK uses WithAuthToken()
+		// and clear the centralized key. If X-Api-Key is present it means the user
+		// has a personal API key; overwrite the centralized key with it.
+		authHeaderName := p.AuthHeader()
+		if bearer := r.Header.Get("Authorization"); bearer != "" {
+			cfg.BYOKBearerToken = strings.TrimPrefix(bearer, "Bearer ")
```
ssncferreira: We should at least check if this is empty. Copilot provider already has an `extractBearerToken` method, maybe we can extract that to a helper and use it here.

provider/anthropic.go
ssncferreira: > I’ll address it in the next PR

Sounds good to me, but can you mention that in the PR description? 🙂

provider/anthropic_test.go
```diff
@@ -105,11 +104,77 @@ func TestAnthropic_CreateInterceptor(t *testing.T) {
 		// Verify the full Anthropic-Beta header (all betas) was forwarded unchanged.
 		assert.Equal(t, betaHeader, receivedHeaders.Get("Anthropic-Beta"), "Anthropic-Beta header must be forwarded unchanged to upstream")
 
-		// Verify aibridge's configured key was used and the client's auth credential was not forwarded.
-		assert.Equal(t, "test-key", receivedHeaders.Get("X-Api-Key"), "upstream must receive configured provider key")
-		assert.Empty(t, receivedHeaders.Get("Authorization"), "client Authorization header must not reach upstream")
+		// The client sent Authorization: Bearer, so BYOK bearer mode is active.
+		// The SDK uses Authorization (not X-Api-Key) for bearer auth.
```
ssncferreira: Not sure this comment is correct, we pass `authHeaderName` as `Authorization` header, so then client_headers properly sets this in `BuildUpstreamHeaders`, right?

provider/anthropic_test.go
```diff
@@ -124,6 +189,52 @@ func TestAnthropic_CreateInterceptor(t *testing.T) {
 	})
 }
 
+func TestAnthropic_InjectAuthHeader_BYOK(t *testing.T) {
```
ssncferreira: Nit: not just for the BYOK case I believe
```suggestion
func TestAnthropic_InjectAuthHeader(t *testing.T) {
```

intercept/messages/base.go
dannykopping: What I mean is: this log message will be disconnected from actual usage; you can't associate this log line with an individual or a request, so it'll be difficult to correlate.

passthrough.go
ssncferreira: This will show the github token for the passthrough routes on Copilot...it should still be masked, but not sure if we should do that.

provider/anthropic.go
evgeniy-scherbina: From my manual testing - it never happens, but I added test-case anyway.
I decided that `X-Api-Key` overrides `Authz`, it's in line with claude behavior:
https://code.claude.com/docs/en/env-vars, see `ANTHROPIC_API_KEY` docs.

provider/anthropic_test.go
evgeniy-scherbina: done

provider/anthropic_test.go
evgeniy-scherbina: I changed the naming of all `TestAnthropic_InjectAuthHeader_BYOK` subtests to be consistent.

provider/anthropic_test.go
evgeniy-scherbina: I added test. I decided let's forward both in this case. It shouldn't happen, but if for some reason claude will set both headers, let's forward both.

utils/mask.go
evgeniy-scherbina: @dannykopping it was explicitly mentioned in PRD and in RFC.

Anthropic in it's console shows 20 chars, e.g.:
`sk-ant-api00-aaa...zzzz`
See: https://platform.claude.com/settings/keys

Since we are provider-agnostic and it’s possible to have very short keys where even 8 characters matter, I propose the following changes:
- log only the last 4 characters
- if the key length is ≤ 12, do not log it at all; return ***

what do you think?

cc @ssncferreira

utils/mask_test.go
evgeniy-scherbina: answered above

passthrough.go
evgeniy-scherbina: In any case, I’m going to update the MaskSecret function to reveal less data, as described above.

@ssncferreira, logging masked OAuth tokens doesn’t provide much value compared to logging masked API keys, so I can remove it from all places?

provider/anthropic.go
evgeniy-scherbina: Definitely make sense, I moved `extractBearerToken` to `utils/auth.go` and reused it. Moved its tests as well.

provider/anthropic_test.go
evgeniy-scherbina: I updated this test along with its comments.

As I understand it, the original purpose of the test was:
- to ensure the `Anthropic-Beta` header is preserved  
- to ensure the `Authorization` header is stripped  

However, the behavior has changed. Now, if `Authorization` is present, it is treated as a BYOK scenario and the header is forwarded.

I adjusted the test accordingly. Now its purpose is:
- to ensure the `Anthropic-Beta` header is preserved (unchanged)  
- when both `Authorization` and `X-Api-Key` are set, to ensure only `X-Api-Key` is forwarded, since `authHeaderName = X-Api-Key` (i.e., `X-Api-Key` has higher priority)

This means `Messages_ClientHeaders` is now indirectly testing a different part of the code (`X-Api-Key` vs `Authorization` priority). If that’s not desirable, I can revise the test—for example, by validating that `Proxy-Authorization` is not forwarded instead. This way, the test won’t fail if the priority between `X-Api-Key` and `Authorization` changes, while still serving its original purpose.

provider/anthropic_test.go
evgeniy-scherbina: This test was added mostly for BYOK, but technically you're right.

provider/anthropic.go
evgeniy-scherbina: I added `Scope` section which mentions it.

config/config.go
evgeniy-scherbina: done

intercept/messages/base.go
evgeniy-scherbina: I'll update this

config/config.go
```diff
@@ -15,6 +15,10 @@ type Anthropic struct {
 	CircuitBreaker   *CircuitBreaker
 	SendActorHeaders bool
 	ExtraHeaders     map[string]string
+	// BYOKBearerToken is set in BYOK mode when the user authenticates
+	// with an OAuth token. When set, the OAuth token is used for upstream
```
dannykopping: Nit: not necessarily OAuth.
Please change elsewhere as well.

```suggestion
	// with a provider token. When set, the provider token is used for upstream
```

provider/anthropic.go
```diff
@@ -110,11 +111,33 @@ func (p *Anthropic) CreateInterceptor(w http.ResponseWriter, r *http.Request, tr
...
+		//
+		// In centralized mode neither Authorization nor X-Api-Key is
+		// present, so cfg keeps the centralized key unchanged.
+		//
+		// In BYOK mode the user's LLM credentials survive intact.
+		// If X-Api-Key is present the user has a personal API key;
+		// overwrite the centralized key with it. If Authorization is
+		// present the user authenticated via OAuth; set BYOKBearerToken
+		// and clear the centralized key. When both are present,
+		// X-Api-Key takes priority.
```
dannykopping: Was this deliberate, or is there a reason to prefer this one?
I worry that if both are provided then it won't be easy to reason about which one is used; we might need to fail with a `400 Bad Request`.

utils/mask.go
dannykopping: Indeed; this was mentioned in the [PRD](https://www.notion.so/coderhq/BYOK-for-AI-Bridge-312d579be59281a1ad85c7d87f19c8ef?source=copy_link#314d579be59280caa652f0c481b04e60).

utils/mask.go
dannykopping: > @dannykopping it was explicitly mentioned in PRD and in RFC.

What do you mean? That we should log the whole key if it's shorter than 8 chars?

---

Ideally we'd log the first few chars because that identifies the key, e.g. `sk-ant` for Anthropic, `sk-proj` for OpenAI, etc so only logging the last 4 might make this slightly less useful.

Ideally we want to:
1. Preserve the length
2. Preserve as much identifying information as we can without leaking a significant portion of the key
3. Be mindful of multi-byte UTF-8 chars (emoji); naïve slice-based splitting will produce garbage

Here's what I propose: https://go.dev/play/p/vVuHboWI1sB

evgeniy-scherbina: @dannykopping
> What do you mean? That we should log the whole key if it's shorter than 8 chars?

Usually keys/tokens should be at least 20 bytes (28 base64 characters), so my reasoning was that exposing 8 out of 28 characters is acceptable.
That said, I agree we should tighten this further.

evgeniy-scherbina: @dannykopping your implementation is reasonable, I used it.
My only concern that OAuth tokens can be very long, maybe it make sense to limit number of `*` symbols, e.g. no more than 50? To avoid very long masked strings?

dannykopping: 👍

I'm fine with making the masked secrets a fixed length, but then we should just also log the unmasked length along with it.

evgeniy-scherbina: @dannykopping do we want to log/store `oauth` tokens (or anything in `Authz Bearer` header) at all? I think it's less useful because oauth tokens probably short-lived, and users often don't have control how they are generated?

Examples of what can be stored in `Authz Bearer`:
- anthropic/openai subscription OAuth Token
- copilot (github) OAuth Token

provider/anthropic.go
evgeniy-scherbina: It was a deliberate choice to be in line with claude-code, see docs: https://code.claude.com/docs/en/env-vars.
```
API key sent as X-Api-Key header. When set, this key is used instead of your
Claude Pro, Max, Team, or Enterprise subscription even if you are logged in.
```

Also you probably saw this in claude-code before:
```
Auth conflict: Both a token (claude.ai) and an API key (ANTHROPIC_API_KEY) are set.

    • Trying to use claude.ai? Unset the ANTHROPIC_API_KEY environment variable,
     or claude /logout then say "No" to the API key approval before login.
    • Trying to use ANTHROPIC_API_KEY? claude /logout to sign out of claude.ai.
```

If both authn methods are used, claude-code uses `X-Api-Key`.

<hr/>

> I worry that if both are provided then it won't be easy to reason about which one is used; we might need to fail with a 400 Bad Request.

Seems it doesn't happen on practice with claude-code. Also I guess upstream LLM API should return 400 if it's not supported.

utils/mask.go
dannykopping: We can't know apriori what the tokens are, so we should proactively store them in case they're long-lived tokens which need identification.

provider/anthropic.go
dannykopping: OK cool, I'm fine with that. Please add the reasoning to the comment 👍

intercept/messages/base.go
evgeniy-scherbina: @dannykopping 
Regarding this, the issue is that the logger is not enriched inside the interceptor (see: https://github.com/coder/aibridge/blob/main/bridge.go#L194).

Would it be okay to move the logger enrichment step (https://github.com/coder/aibridge/blob/main/bridge.go#L213-L220) before calling `interceptor.Setup(logger, asyncRecorder, mcpProxy)`?

<hr/>

Currently log record looks like this:
```
coderd.aibridged.pool.streaming:
using byok oauth bearer auth 
request_id=fdafd5e4-d06b-4bac-8949-4dede8280c0d
bearer_hint=sk-a...ABCD
```

After logger enrichment it will look like this:
```
coderd.aibridged.pool.streaming:
using byok oauth bearer auth
route=/v1/messages
provider=anthropic
interception_id=cda0a3ae-b187-463e-9cb8-a5601519e785
user_agent="claude-cli/2.1.81 (external, cli)"
streaming=true
request_id=fb990efc-0a58-4c8c-b32b-fce070745664
bearer_hint=sk-a...ABCD
```

evgeniy-scherbina: Alternatively, as mentioned earlier, I can move the logs into `Anthropic.CreateInterceptor` and include the `interceptionID` in the logs.

ssncferreira: Also agree that the logs here don't add much value, would prefer to see them in the context of an interception, especially on error logs related to auth failures. But I'm ok with addressing this in a follow up issue in case it increases the scope of this PR a lot.

provider/anthropic_test.go
```diff
@@ -105,11 +106,84 @@ func TestAnthropic_CreateInterceptor(t *testing.T) {
...
 		assert.Equal(t, betaHeader, receivedHeaders.Get("Anthropic-Beta"), "Anthropic-Beta header must be forwarded unchanged to upstream")
 
-		// Verify aibridge's configured key was used and the client's auth credential was not forwarded.
-		assert.Equal(t, "test-key", receivedHeaders.Get("X-Api-Key"), "upstream must receive configured provider key")
+		// Verify user's personal key was used and the authorization header was not forwarded.
+		assert.Equal(t, "personal user key", receivedHeaders.Get("X-Api-Key"), "upstream must receive personal user key")
 		assert.Empty(t, receivedHeaders.Get("Authorization"), "client Authorization header must not reach upstream")
 	})
 
+	byokTests := []struct {
```
ssncferreira: Nit: I would suggest moving these tests to their own test function, just to make it clearer

passthrough.go
evgeniy-scherbina: `UPD`:
@dannykopping good catch - fixed. I reused `utils.ExtractBearerToken`.
@ssncferreira according to Danny's comments above, we want to store them.

utils/mask_test.go
```diff
@@ -0,0 +1,32 @@
...
+		name     string
+		input    string
+		expected string
+	}{
+		{"empty", "", ""},
+		{"short", "short", "*****"},
+		{"short_9_chars", "veryshort", "*********"},
+		{"medium_15_chars", "thisisquitelong", "th***********ng"},
+		{"long_api_key", "sk-ant-api03-abcdefgh", "sk-a*************efgh"},
+		{"unicode", "hélloworld🌍!", "hé********🌍!"},
```
ssncferreira, line 23: Nit: I would suggest adding a github token subtest as well, just to validate that it is successfully masked: https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/about-authentication-to-github#githubs-token-formats

utils/mask.go
```diff
@@ -0,0 +1,37 @@
+package utils
+
+import "strings"
+
+// MaskSecret masks the middle of a secret string, revealing a small
+// prefix and suffix for identification. The number of characters
+// revealed scales with string length.
+func MaskSecret(s string) string {
```
ssncferreira, line 8: I just remembered that we have https://github.com/coder/aibridge/blob/main/intercept/apidump/headers.go#L25
I believe it handles this already properly, no? Maybe we can extract that here, and reuse it for apidump as well.

passthrough.go
```diff
@@ -96,6 +97,16 @@ func newPassthroughRouter(provider provider.Provider, logger slog.Logger, m *met
 
 				// Inject provider auth.
 				provider.InjectAuthHeader(&req.Header)
+
+				if apiKey := req.Header.Get("X-Api-Key"); apiKey != "" {
+					logger.Debug(ctx, "passthrough using api key auth",
+						slog.F("api_key_hint", utils.MaskSecret(apiKey)),
+					)
+				} else if token := utils.ExtractBearerToken(req.Header.Get("Authorization")); token != "" {
+					logger.Debug(ctx, "passthrough using oauth bearer auth",
```
ssncferreira: I think adding this will generate a lot of unnecessary noise, so I would suggest removing it and focusing on the logs in the context of interceptions

ssncferreira (commented):
A few comments, but I think it is looking good 👍

passthrough.go
evgeniy-scherbina: `UPD2`:
Removed it as per latest discussion with @ssncferreira

intercept/messages/base.go
dannykopping: @evgeniy-scherbina that looks great 👍 go ahead
Agree with Susana that if you wanna do it in a follow-up that's fine

utils/mask.go
dannykopping: Loooool I _knew_ I wrote that somewhere but couldn't remember where. Thank you!
@evgeniy-scherbina's implementation is better though.

intercept/messages/base.go
evgeniy-scherbina: okay, I'll do in a follow-up

dannykopping (approved):
LGTM

If you want to do the logging improvements in a follow-up, feel free to merge this one and move on 👍 

Great work!

config/config.go
evgeniy-scherbina: Replaced everywhere I used `access token` instead of `oauth token`.

provider/anthropic.go
evgeniy-scherbina: done

provider/anthropic_test.go
evgeniy-scherbina: done

utils/mask_test.go
evgeniy-scherbina: Done for `ghp_` prefix.

utils/mask.go
evgeniy-scherbina: I'll do this in a follow-up PR.

passthrough.go
evgeniy-scherbina: Done

utils/mask.go
evgeniy-scherbina: Addressed here: https://github.com/coder/aibridge/pull/251

PR #218, opened 2026-03-16
chore: remove SDK struct usage from requests in messages API

pawbana:
Replaces MessageNewParamsWrapper that wrapped anthropic.MessageNewParams
with MessagesRequestPayload that operates on raw bytes.

PR #215, opened 2026-03-12
chore: move lastUserPrompt and correlatingToolCallID to ResponsesRequestPayload

pawbana:
Moved `correlatingToolCallID()` and `lastUserPrompt()` methods from `responsesInterceptionBase` to `ResponsesRequestPayload`

PR #213, opened 2026-03-12
chore: replace ResponsesNewParamsWrapper with ResponsesRequestPayload in responses interceptor

pawbana:
Refactored the responses interceptor replacing `ResponsesNewParamsWrapper` with a new `ResponsesRequestPayload` type that wraps raw JSON bytes and consolidates operations made on request payload.

PR #225, opened 2026-03-19
fix: remove unsupported fields from requests to Bedrock

### LINKED ISSUE #219: bug: Bedrock models that don't support adaptive thinking reject `thinking.type: "adaptive"` requests
## Summary

When using AI Bridge with Amazon Bedrock in regions where only older Claude models are available (e.g. GovCloud `us-gov-west-1`), requests from Claude Code fail with a 400 error because the `thinking` parameter is passed through unmodified to models that don't support `"type": "adaptive"`.

## Error

```
API Error: 400 {"error":{"message":"POST \"https://bedrock-runtime.us-gov-west-1.amazonaws.com/v1/messages\": 400 Bad Request {\"message\":\"thinking: Input tag 'adaptive' found using 'type' does not match any of the expected tags: 'disabled', 'enabled'\"}","type":"api_error"},"request_id":"","type":"error"}
```

## Root Cause

Claude Code now sends `"thinking": {"type": "adaptive"}` by default (introduced with Opus 4.6 / Sonnet 4.6). The `augmentRequestForBedrock()` function in `intercept/messages/base.go` only remaps the model name — it does **not** transform or validate the `thinking` parameter for the target model's capabilities.

In GovCloud, the most advanced FedRAMP-authorized Claude model is **Claude Sonnet 4.5**, which only supports `"thinking": {"type": "enabled", "budget_tokens": N}` or `{"type": "disabled"}`. The `"adaptive"` type is only supported by Opus 4.6 and Sonnet 4.6.

## Expected Behavior

`augmentRequestForBedrock()` should transform the `thinking` parameter based on the target Bedrock model's capabilities:

* If the model supports adaptive thinking (Opus 4.6, Sonnet 4.6): pass through as-is
* Otherwise: convert `"adaptive"` to `"enabled"` with a sensible `budget_tokens` default, or fall back to the model's supported thinking configuration

## Impact

This is currently blocking usage of Claude Code through AI Bridge with Bedrock in GovCloud regions (and any other Bedrock deployment using models that predate adaptive thinking support).

## References

* [Anthropic Adaptive Thinking docs](<https://platform.claude.com/docs/en/build-with-claude/extended-thinking>)
* [AWS Bedrock Adaptive Thinking docs](<https://docs.aws.amazon.com/bedrock/latest/userguide/claude-messages-adaptive-thinking.html>)
* [AWS GovCloud Bedrock model availability](<https://docs.aws.amazon.com/govcloud-us/latest/UserGuide/govcloud-bedrock.html>)
* Related prior fix for adaptive thinking being stripped: coder/aibridge#185 / [AIGOV-19](https://linear.app/codercom/issue/AIGOV-19/claude-opus-46-support-thinking-configuration-and-json-schema)

---

Created on behalf of @uzair-coder07

pawbana:
This PR removes unsupported fields from requests to Bedrock.
Adaptive thinking field is replaced for models that don't support with `type: enabled` + `budget_tokens: N` combination.
`budget_tokens` value is calculated as a ratio of `max_tokens`.
Ratio depends on `output_config.effort` value, used the same ratio values as OpenRouter does: https://openrouter.ai/docs/guides/best-practices/reasoning-tokens#reasoning-effort-level

Fixes: https://github.com/coder/aibridge/issues/219

### DIFF
--- a/intercept/messages/base.go
+++ b/intercept/messages/base.go
@@ -277,19 +277,45 @@ func (i *interceptionBase) withAWSBedrockOptions(ctx context.Context, cfg *aibco
 }
 
 // augmentRequestForBedrock will change the model used for the request since AWS Bedrock doesn't support
-// Anthropics' model names.
+// Anthropics' model names. It also converts adaptive thinking to enabled with a budget for models that
+// don't support adaptive thinking natively.
 func (i *interceptionBase) augmentRequestForBedrock() {
 	if i.bedrockCfg == nil {
 		return
 	}
 
-	updated, err := i.reqPayload.withModel(i.Model())
+	model := i.Model()
+	updated, err := i.reqPayload.withModel(model)
 	if err != nil {
 		i.logger.Warn(context.Background(), "failed to set model in request payload for Bedrock", slog.Error(err))
 		return
 	}
-
 	i.reqPayload = updated
+
+	if !bedrockModelSupportsAdaptiveThinking(model) {
+		updated, err = i.reqPayload.convertAdaptiveThinkingForBedrock()
+		if err != nil {
+			i.logger.Warn(context.Background(), "failed to convert adaptive thinking for Bedrock", slog.Error(err))
+			return
+		}
+		i.reqPayload = updated
+	}
+
+	// Strip fields that Bedrock does not accept.
+	updated, err = i.reqPayload.removeUnsupportedBedrockFields()
+	if err != nil {
+		i.logger.Warn(context.Background(), "failed to remove unsupported fields for Bedrock", slog.Error(err))
+		return
+	}
+	i.reqPayload = updated
+}
+
+// bedrockModelSupportsAdaptiveThinking returns true if the given Bedrock model ID
+// supports the "adaptive" thinking type natively (i.e. Claude 4.6 models).
+// See https://docs.aws.amazon.com/bedrock/latest/userguide/claude-messages-adaptive-thinking.html
+func bedrockModelSupportsAdaptiveThinking(model string) bool {
+	return strings.Contains(model, "anthropic.claude-opus-4-6") ||
+		strings.Contains(model, "anthropic.claude-sonnet-4-6")
 }
 
 // writeUpstreamError marshals and writes a given error.
--- a/intercept/messages/reqpayload.go
+++ b/intercept/messages/reqpayload.go
@@ -14,8 +14,19 @@ import (
 const (
 	// Absolute JSON paths from the request root.
 	messagesReqPathMessages                  = "messages"
+	messagesReqPathMaxTokens                 = "max_tokens"
 	messagesReqPathModel                     = "model"
+	messagesReqPathOutputConfig              = "output_config"
+	messagesReqPathOutputConfigEffort        = "output_config.effort"
+	messagesReqPathMetadata                  = "metadata"
+	messagesReqPathServiceTier               = "service_tier"
+	messagesReqPathContainer                 = "container"
+	messagesReqPathInferenceGeo              = "inference_geo"
+	messagesReqPathContextManagement         = "context_management"
 	messagesReqPathStream                    = "stream"
+	messagesReqPathThinking                  = "thinking"
+	messagesReqPathThinkingBudgetTokens      = "thinking.budget_tokens"
+	messagesReqPathThinkingType              = "thinking.type"
 	messagesReqPathToolChoice                = "tool_choice"
 	messagesReqPathToolChoiceDisableParallel = "tool_choice.disable_parallel_tool_use"
 	messagesReqPathToolChoiceType            = "tool_choice.type"
@@ -29,6 +40,12 @@ const (
 	messagesReqFieldType      = "type"
 )
 
+const (
+	constAdaptive = "adaptive"
+	constDisabled = "disabled"
+	constEnabled  = "enabled"
+)
+
 var (
 	constAny        = string(constant.ValueOf[constant.Any]())
 	constAuto       = string(constant.ValueOf[constant.Auto]())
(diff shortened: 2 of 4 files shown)


pawbana:
* **#226** <a href="https://app.graphite.com/github/pr/coder/aibridge/226?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#225** <a href="https://app.graphite.com/github/pr/coder/aibridge/225?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/aibridge/225?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#218** <a href="https://app.graphite.com/github/pr/coder/aibridge/218?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#215** <a href="https://app.graphite.com/github/pr/coder/aibridge/215?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#213** <a href="https://app.graphite.com/github/pr/coder/aibridge/213?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

intercept/messages/base.go
```diff
@@ -264,21 +264,47 @@ func (i *interceptionBase) withAWSBedrockOptions(ctx context.Context, cfg *aibco
...
+		return
+	}
 	i.reqPayload = updated
 }
 
+// bedrockModelSupportsAdaptiveThinking returns true if the given Bedrock model ID
+// supports the "adaptive" thinking type natively (i.e. Claude 4.6 models).
+// See https://docs.aws.amazon.com/bedrock/latest/userguide/claude-messages-adaptive-thinking.html
+func bedrockModelSupportsAdaptiveThinking(model string) bool {
+	return strings.Contains(model, "anthropic.claude-opus-4-6") ||
```
dannykopping, line 317: This doesn't match what the docs say:

```suggestion
	return strings.Contains(model, "anthropic.claude-opus-4-6-v1") ||
```

intercept/messages/reqpayload.go
```diff
@@ -265,7 +297,71 @@ func (p MessagesRequestPayload) tools() ([]json.RawMessage, error) {
...
+
+	maxTokens := gjson.GetBytes(p, messagesReqPathMaxTokens).Int()
+	if maxTokens <= 0 {
+		// max_tokens is required by messages API
+		return p, fmt.Errorf("max_tokens: field required")
+	}
+
+	effort := gjson.GetBytes(p, messagesReqPathOutputConfigEffort).String()
+
+	// Effort-to-ratio mapping adapted from OpenRouter:
```
dannykopping, line 334: This explains the what, but not the why.

Can you please add more detail about why we're augmenting this value?

intercept/messages/reqpayload.go
```diff
@@ -265,7 +297,71 @@ func (p MessagesRequestPayload) tools() ([]json.RawMessage, error) {
...
+}
+
 func (p MessagesRequestPayload) set(path string, value any) (MessagesRequestPayload, error) {
 	out, err := sjson.SetBytes(p, path, value)
 	return MessagesRequestPayload(out), err
 }
+
+func (p MessagesRequestPayload) delete(path string) (MessagesRequestPayload, error) {
+	out, err := sjson.DeleteBytes(p, path)
+	return MessagesRequestPayload(out), err
```
dannykopping: Please wrap error.

intercept/messages/reqpayload.go
```diff
@@ -265,7 +297,71 @@ func (p MessagesRequestPayload) tools() ([]json.RawMessage, error) {
...
+
+// removeUnsupportedBedrockFields strips all top-level fields that Bedrock does
+// not support from the payload.
+func (p MessagesRequestPayload) removeUnsupportedBedrockFields() (MessagesRequestPayload, error) {
+	result := p
+	for _, field := range bedrockUnsupportedFields {
+		var err error
+		result, err = result.delete(field)
+		if err != nil {
+			return p, fmt.Errorf("removing %q: %w", field, err)
```
dannykopping: This isn't enough explanatory detail.

```suggestion
			return p, fmt.Errorf("removing %q field for Bedrock compatibility: %w", field, err)
```

intercept/messages/reqpayload.go
```diff
@@ -265,7 +297,71 @@ func (p MessagesRequestPayload) tools() ([]json.RawMessage, error) {
...
+	})
+}
+
+// removeUnsupportedBedrockFields strips all top-level fields that Bedrock does
+// not support from the payload.
+func (p MessagesRequestPayload) removeUnsupportedBedrockFields() (MessagesRequestPayload, error) {
+	result := p
+	for _, field := range bedrockUnsupportedFields {
+		var err error
+		result, err = result.delete(field)
```
dannykopping: Nit: doing this once per field means we have to parse the JSON each time.

Probably not a huge problem but I worry about memory usage since this will spill to the heap.  
You could consider unmarshaling once into a map, deleting there, and remarshaling.

intercept/messages/base.go
pawbana: I've went with more permissive check in case new Bedrock version of `opus-4-6` would be released.
It is impossible to know how exactly how new version would behave but I'd assume it should support the same fields / beta flags (or more). I'm fine with being more restrictive.

intercept/messages/reqpayload.go
pawbana: Good point. I think it is simple enough to include this change here.  
  
There are few other places where performance could be improved (eg. injecting tools / appending messages in requests). To fix it properly I think we need some map like struct (eg. https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured) but it is out of scope. Also most improvements may not be needed after deprecation of MCP injection.

intercept/messages/reqpayload.go
pawbana: with map re-marshalling change this error is no longer possible

intercept/messages/reqpayload.go
pawbana: Extended comments.

dannykopping (approved):

pawbana:
### Merge activity

* **Mar 26, 4:28 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/coder/aibridge/225).
* **Mar 26, 4:30 PM UTC**: [Graphite](https://app.graphite.com/github/pr/coder/aibridge/225) rebased this pull request as part of a merge.
* **Mar 26, 4:33 PM UTC**: @pawbana merged this pull request with [Graphite](https://app.graphite.com/github/pr/coder/aibridge/225).

PR #227, opened 2026-03-20
feat: aibrige BYOK for OpenAI

evgeniy-scherbina:
### Overview
   
`CreateInterceptor` detects user LLM credentials from request headers and sets them on the config copy. `InjectAuthHeader` skips centralized key injection when user credentials are already present (passthrough BYOK).                                                           
                                                                                                                                                      
  ### End-to-end flow                                                                                                              
                                                                                                                                                      
  **Centralized:** No user credentials in the request → centralized key is used.                                                            
   
  **BYOK (personal api key):** User's credential set in `Authorization`, `CreateInterceptor` detects it and replaces the centralized key with it.                                                                                     
                                                                                                                                                      
  ~~**ChatGPT subscription (Plus/Pro):** Same as BYOK (personal api key) but `Chatgpt-Account-Id` triggers routing to `chatgpt.com/backend-api/codex` instead of `api.openai.com/v1`, since subscription OAuth tokens are only accepted by the ChatGPT-specific backend.~~
                                                                   
   ### Related PRs                                                                                                                                                                                                                           
  Mirrors the Anthropic BYOK implementation from #216.

### DIFF
--- a/provider/openai.go
+++ b/provider/openai.go
@@ -104,7 +104,7 @@ func (p *OpenAI) CreateInterceptor(w http.ResponseWriter, r *http.Request, trace
 	// In centralized mode Authorization is absent, so cfg keeps the
 	// centralized key unchanged.
 	//
-	// In BYOK mode the user's bearer token is in Authorization. Replace
+	// In BYOK mode the user's credential is in Authorization. Replace
 	// the centralized key with it so the SDK forwards it upstream.
 	if token := utils.ExtractBearerToken(r.Header.Get("Authorization")); token != "" {
 		cfg.Key = token
--- a/provider/openai.go
+++ b/provider/openai.go
@@ -22,8 +22,6 @@ import (
 const (
 	routeChatCompletions = "/chat/completions" // https://platform.openai.com/docs/api-reference/chat
 	routeResponses       = "/responses"        // https://platform.openai.com/docs/api-reference/responses
-
-	chatGPTBaseURL = "https://chatgpt.com/backend-api/codex"
 )
 
 var openAIOpenErrorResponse = func() []byte {
@@ -110,14 +108,6 @@ func (p *OpenAI) CreateInterceptor(w http.ResponseWriter, r *http.Request, trace
 	// the centralized key with it so it is forwarded upstream.
 	if token := utils.ExtractBearerToken(r.Header.Get("Authorization")); token != "" {
 		cfg.Key = token
-	}
-
-	// ChatGPT subscription OAuth tokens (Plus/Pro) use a different backend
-	// than platform API keys. The Chatgpt-Account-Id header is set by Codex
-	// when authenticating via a ChatGPT subscription, indicating that requests
-	// must be routed to the subscription-specific backend.
-	if r.Header.Get("Chatgpt-Account-Id") != "" {
-		cfg.BaseURL = chatGPTBaseURL
 	}
 
 	path := strings.TrimPrefix(r.URL.Path, p.RoutePrefix())
--- a/provider/openai.go
+++ b/provider/openai.go
@@ -22,6 +22,8 @@ import (
 const (
 	routeChatCompletions = "/chat/completions" // https://platform.openai.com/docs/api-reference/chat
 	routeResponses       = "/responses"        // https://platform.openai.com/docs/api-reference/responses
+
+	chatGPTBaseURL = "https://chatgpt.com/backend-api/codex"
 )
 
 var openAIOpenErrorResponse = func() []byte {
@@ -105,7 +107,7 @@ func (p *OpenAI) CreateInterceptor(w http.ResponseWriter, r *http.Request, trace
 	// centralized key unchanged.
 	//
 	// In BYOK mode the user's credential is in Authorization. Replace
-	// the centralized key with it so the SDK forwards it upstream.
+	// the centralized key with it so it is forwarded upstream.
 	if token := utils.ExtractBearerToken(r.Header.Get("Authorization")); token != "" {
 		cfg.Key = token
 	}
@@ -115,7 +117,7 @@ func (p *OpenAI) CreateInterceptor(w http.ResponseWriter, r *http.Request, trace
 	// when authenticating via a ChatGPT subscription, indicating that requests
 	// must be routed to the subscription-specific backend.
 	if r.Header.Get("Chatgpt-Account-Id") != "" {
-		cfg.BaseURL = "https://chatgpt.com/backend-api/codex"
+		cfg.BaseURL = chatGPTBaseURL
 	}
 
 	path := strings.TrimPrefix(r.URL.Path, p.RoutePrefix())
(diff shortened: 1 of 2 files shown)


provider/openai.go
```diff
@@ -96,6 +99,27 @@ func (p *OpenAI) CreateInterceptor(w http.ResponseWriter, r *http.Request, trace
...
+	}
+
+	// ChatGPT subscription OAuth tokens (Plus/Pro) use a different backend
+	// than platform API keys. The Chatgpt-Account-Id header is set by Codex
+	// when authenticating via a ChatGPT subscription, indicating that requests
+	// must be routed to the subscription-specific backend.
+	if r.Header.Get("Chatgpt-Account-Id") != "" {
+		cfg.BaseURL = chatGPTBaseURL
+	}
+
```
evgeniy-scherbina: `NOTE`: I didn't add test for this specific logic, I want to agree on solution before doing it.

<hr/>

Difference between API-key based request and ChatGPT subscription-based request:

| Header | `api.openai.com` (API key) | `chatgpt.com` (subscription) |                                                                              
  |--------|---------------------------|-------------------------------|                                                                            
  | `Host` | `api.openai.com` | `chatgpt.com` |                                                                                                       
  | `Path` | `/v1/responses` | `/backend-api/codex/responses` |                                                                                       
  | `Authorization` | `Bearer sk-proj-...` (API key) | `Bearer eyJ...` (JWT OAuth) |                                                                  
  | `Chatgpt-Account-Id` | absent | present | 

<hr/>

Overview of different approaches:

| Option | Reliability | Notes |                                                                                                                    
  |---|---|---|                                                                                                                                       
  | `Chatgpt-Account-Id` header | High | Already implemented; Codex sets it specifically for subscription auth |                                      
  | `IsJWT` on `Authorization` | Medium | ChatGPT tokens are JWTs (`eyJ...`); API keys are `sk-proj-...`. Works today but fragile — OpenAI could add  JWT-based API keys |                                                                                                                                
  | Custom Coder header | High | Requires users to configure Codex to send an extra header indicating they are using ChatGPT subscription mode |

> [!NOTE]                                                                                                                                           
   Host and path information from the original request is not available to aibridge. We set
   `base_url = "https://host.coder.com/api/v2/aibridge/openai/v1"`, so aibridge only sees paths like `/responses` or `/chat/completions` — the original upstream host (`chatgpt.com` vs `api.openai.com`) is never forwarded. This is why request routing decisions must be based on headers rather than the original target URL.

provider/openai_test.go
```diff
@@ -163,22 +163,44 @@ func TestOpenAI_CreateInterceptor(t *testing.T) {
...
-			name:         "Responses_ClientHeaders",
-			route:        routeResponses,
-			requestBody:  `{"model": "gpt-5", "input": "hello", "stream": false}`,
-			responseBody: responsesAPIResponse,
+			name:              "ChatCompletions_Centralized_UsesCentralizedKey",
+			route:             routeChatCompletions,
+			requestBody:       `{"model": "gpt-4", "messages": [{"role": "user", "content": "hello"}], "stream": false}`,
+			responseBody:      chatCompletionResponse,
+			setHeaders:        map[string]string{},
+			wantAuthorization: "Bearer test-key",
```
ssncferreira: nit:
```suggestion
			wantAuthorization: "Bearer centralized-key",
```

provider/openai_test.go
```diff
@@ -163,22 +163,44 @@ func TestOpenAI_CreateInterceptor(t *testing.T) {
 	t.Parallel()
 
 	tests := []struct {
-		name         string
-		route        string
-		requestBody  string
-		responseBody string
+		name              string
```
ssncferreira, line 166: I would suggest leaving this for just (maybe focusing on just centralied mode (default):
* `ChatCompletions_ClientHeaders`
* `Responses_ClientHeaders`

And adding the BYOK tests as a separate test function, just to keep it consistent with Anthropic `TestAnthropic_CreateInterceptor_BYOK`

provider/openai_test.go
```diff
@@ -219,13 +241,50 @@ func TestOpenAI_CreateInterceptor(t *testing.T) {
...
+		{
+			name:              "when no Authorization header is provided, inject centralized key",
+			presetHeaders:     map[string]string{},
+			wantAuthorization: "Bearer centralized-key",
+		},
+		{
+			name:              "when Authorization header is provided, do not overwrite it",
+			presetHeaders:     map[string]string{"Authorization": "Bearer user-token"},
+			wantAuthorization: "Bearer user-token",
+		},
```
ssncferreira, line 293: What if `X-API-Key` is present, what happens? Or what is expected to happen?

provider/openai_test.go
```diff
@@ -163,22 +163,44 @@ func TestOpenAI_CreateInterceptor(t *testing.T) {
...
+			setHeaders:        map[string]string{"Authorization": "Bearer user-token"},
+			wantAuthorization: "Bearer user-token",
+		},
+		{
+			name:              "Responses_Centralized_UsesCentralizedKey",
+			route:             routeResponses,
+			requestBody:       `{"model": "gpt-5", "input": "hello", "stream": false}`,
+			responseBody:      responsesAPIResponse,
+			setHeaders:        map[string]string{},
+			wantAuthorization: "Bearer test-key",
```
ssncferreira: Similarly to Anthropic we should add a test for the case both Authorization and X-API-Key are present.

ssncferreira (approved):
Overall LGTM, just a few improvements on the tests

provider/openai_test.go
evgeniy-scherbina: @ssncferreira initially I did it (see first commit: https://github.com/coder/aibridge/pull/227/changes/4a888ef3ff1c47b861683eb916eebe6f1cd180da), but later realized that `TestAnthropic_CreateInterceptor_BYOK` is a superset of `TestAnthropic_CreateInterceptor`, so I removed the latter.

As you can see from https://github.com/coder/aibridge/pull/227/changes/4a888ef3ff1c47b861683eb916eebe6f1cd180da `TestOpenAI_CreateInterceptor_BYOK` covers both `centralized` and `personal api key`, which make sense we want to cover both positive (BYOK triggered) and negative (BYOK not triggered) scenario. In such case `TestOpenAI_CreateInterceptor` becomes redundant.

provider/openai_test.go
evgeniy-scherbina: I think `X-API-Key` is completely ignored in `openai` scenario. So what happens depends solely on `Authorization` header.

provider/openai_test.go
evgeniy-scherbina: You're sure it's necessary? `X-Api-Key` is not mentioned anywhere in `openai` provider?

provider/openai_test.go
evgeniy-scherbina: done

provider/openai_test.go
ssncferreira: But theoretically, it is possible for a request to include both headers, right? What is expected to happen in that case?

evgeniy-scherbina: They should be stripped by `BuildUpstreamHeaders`, I added tests for this behavior.

PR #230, opened 2026-03-26
chore: capture prompts from haiku models

dannykopping:
**Context:**

When using Claude Code with Sonnet or Opus, it'll use the Haiku model for [ancillary actions](https://code.claude.com/docs/en/costs#background-token-usage) that don't require a lot of intelligence. Claude Code calls this model it's "small fast model".

When inference requests are sent, they include a "role" to discriminate between user- and assistant-initiated messages.

Here's an example of Claude Code generating a session title based on the initial human prompt.

<details>
<summary>Example request</summary>

```json
{
  "model": "claude-haiku-4-5-20251001",
  "messages": [
    {
      "role": "user",
      "content": [
        {
          "type": "text",
          "text": "<system-reminder>\nAs you answer the user's questions, you can use the following context:..."
        },
        {
          "type": "text",
          "text": "Thoroughly explore the dbcrypt package in this codebase. I need to understand:\n\n1. What is dbcrypt and what problem does it solve?\n2. What is the architecture - key types, interfaces, and how they compose?\n3. How does encryption/decryption actually work (what cipher, what fields)?\n4. How is it integrated into the broader system (where is it instantiated, what wraps what)?\n5. Key files and their purposes\n6. How are encryption keys managed (rotation, active key, etc.)?\n\nSearch in enterprise/dbcrypt/ and also look for how it's used in coderd/ and enterprise/. Be very thorough - read the main files fully.",
          "cache_control": {
            "type": "ephemeral"
          }
        }
      ]
    }
  ],
  "system": [
    {
      "type": "text",
      "text": "You are Claude Code, Anthropic's official CLI for Claude.",
      "cache_control": {
        "type": "ephemeral"
      }
    },
    {
      "type": "text",
      "text": "You are a file search specialist for Claude Code, Anthropic's official CLI for Claude...",
      "cache_control": {
        "type": "ephemeral"
      }
    }
  ]
}
```

</details>

This request was initiated _by Claude Code_ not by the user; even so, Claude Code specifies `"role": "user"` which muddies the waters because now we can't know if this was actually initiated by a human or not.

---

When we were only displaying disconnected interceptions this was hard to reason about, so we decided to not capture the prompt for Haiku models. Now that we have the session view, however, it's much clearer how the Haiku models produce their prompts so auditors will be able to reason about it a little better. In this case, Claude Code is spawning an `Explore` agent to poke around the codebase.

image

It's still suboptimal that we can't discriminate human- from agent-initiated prompts, but that's a problem with the source data which we can't currently work around.

_Aside: pity we didn't have tests validating the previous behaviour; that's on me._

### DIFF
--- a/fixtures/anthropic/haiku_simple.txtar
+++ b/fixtures/anthropic/haiku_simple.txtar
@@ -1,0 +1,69 @@
+Simple request using a Haiku model (small/fast model).
+Used to validate that prompts are captured for small/fast models like Haiku,
+which Claude Code uses for ancillary tasks (e.g. generating session titles,
+push notification summaries).
+
+-- request --
+{
+  "max_tokens": 256,
+  "messages": [
+    {
+      "content": [
+        {
+          "text": "Chat title: yo\n\nAgent's last message:\nyo, what's up? How can I help?",
+          "type": "text"
+        }
+      ],
+      "role": "user"
+    }
+  ],
+  "model": "claude-haiku-4-5",
+  "system": [
+    {
+      "text": "You are a notification assistant. Given a chat title and the agent's last message, write a single short sentence summarizing what the agent did.",
+      "type": "text"
+    }
+  ]
+}
+
+-- streaming --
+event: message_start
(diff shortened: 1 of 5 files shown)


pawbana (commented):
As much as I can see this is a minimal change that should work and not break anything it would be nice to have some test for it. Maybe small fixture + test case here: https://github.com/coder/aibridge/blob/dd8c239e55660e69eed19070487cc5743a78ba48/internal/integrationtest/bridge_test.go#L480

pawbana (approved):

PR #232, opened 2026-03-27
feat: chatgpt provider

evgeniy-scherbina:
Related issue:
- https://github.com/coder/internal/issues/1401

Adds a new `ChatGPT` provider that targets the ChatGPT backend. The implementation is based on the OpenAI provider since the backends are API-compatible, but follows the Copilot approach for authentication — credentials are passed per-request via the `Authorization` header rather than configured as a centralized API key.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -6,6 +6,7 @@ const (
 	ProviderAnthropic = "anthropic"
 	ProviderOpenAI    = "openai"
 	ProviderCopilot   = "copilot"
+	ProviderChatGPT   = "chatgpt"
 )
 
 type Anthropic struct {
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -40,6 +41,23 @@ type OpenAI struct {
...
+// Since it authenticates exclusively via per-user credentials, it does not
+// require a centralized API key.
+type ChatGPT struct {
+	BaseURL          string
+	APIDumpDir       string
+	CircuitBreaker   *CircuitBreaker
+	SendActorHeaders bool
+	ExtraHeaders     map[string]string
+}
+
```
evgeniy-scherbina, line 60: Based on my testing, the ChatGPT backend does not support API keys, so there’s no point in supporting a centralized key for the ChatGPT provider.

provider/chatgpt.go
```diff
@@ -0,0 +1,168 @@
...
+
+	var interceptor intercept.Interceptor
+
+	// Extract the per-user ChatGPT token from the Authorization header.
+	token := utils.ExtractBearerToken(r.Header.Get("Authorization"))
+	if token == "" {
+		span.SetStatus(codes.Error, "missing authorization")
+		return nil, fmt.Errorf("missing ChatGPT authorization: Authorization header not found or invalid")
+	}
+
```
evgeniy-scherbina, line 102: If the `Authorization` header is not provided, fail early, mirroring Copilot’s behavior.

evgeniy-scherbina:
We decided to implement another approach.

PR #237, opened 2026-03-30
feat: add Type() to Provider interface and configurable instance Name

ssncferreira:
## Problem

There is no way to register multiple provider instances of the same type with distinct identities. For example, introducing a copilot-business provider (targeting `api.business.githubcopilot.com`) alongside the default copilot provider (targeting `api.individual.githubcopilot.com`) would result in both sharing the same name, making them indistinguishable in metrics, logs, circuit breaker, API dump, database records, and the coder aibridge HTTP route prefix.

## Changes

* Add `Type()` to `Provider` interface returning the provider type ("anthropic", "openai", "copilot")
* Add configurable `Name` field to provider configs, defaulting to the type when not set
* `Name()` now returns the instance name, used for metrics, logs, circuit breaker, routing, and the coder aibridge endpoint (`RoutePrefix()` is derived from `Name()`)
* `Type()` is used for the database `Provider` field
* Add `ProviderName` field to `InterceptionRecord` for the instance name

Related to: https://github.com/coder/aibridge/issues/152

_Disclaimer: initially produced by Claude Opus 4.6, heavily modified and reviewed by @ssncferreira ._


ssncferreira:
* **#240** <a href="https://app.graphite.com/github/pr/coder/aibridge/240?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#238** <a href="https://app.graphite.com/github/pr/coder/aibridge/238?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#237** <a href="https://app.graphite.com/github/pr/coder/aibridge/237?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/aibridge/237?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

bridge.go
```diff
@@ -198,7 +198,8 @@ func newInterceptionProcessor(p provider.Provider, cbs *circuitbreaker.ProviderC
 			InitiatorID:           actor.ID,
 			Metadata:              actor.Metadata,
 			Model:                 interceptor.Model(),
-			Provider:              p.Name(),
+			Provider:              p.Type(),
+			ProviderName:          p.Name(),
```
ssncferreira, line 202: Keeping the provider type ("anthropic", "openai", "copilot") so interceptions preserve this value and we are still easily able to aggregate by provider. The `ProviderName` field provides the specific instance identity. Alternatively, we could store the upstream URL instead.

Examples:
* Anthropic: Provider: anthropic, ProviderName: anthropic
* Copilot: Provider: copilot, ProviderName: copilot
* Copilot Business: Provider: copilot, ProviderName: copilot-business
* Copilot Enterprise: Provider: copilot, ProviderName: copilot-enterprise

config/config.go
```diff
@@ -9,6 +9,8 @@ const (
 )
 
 type Anthropic struct {
+	// Name is the provider instance name. If empty, defaults to ProviderAnthropic.
```
pawbana: nit: comment is a bit confusing, defaults to `"ProviderAnthropic"` or value of `ProviderAnthropic` variable?

Maybe functions `Name()`and `Type()` with logic regarding defaults should be in `config` package and `provider` would simply call them? Then it would be simpler to understand the comment.  
  
Same confusion in comments for other providers.

pawbana:
> `RoutePrefix()` is derived from `Name()`

I'm not sure how I fell about this. If `Name` is configurable I think route prefix should also be configurable but maybe default could be based on `Name`to simplify configuration / UX.

If `RoutePrefix` would be configurable it needs good explanation how it works. Due to how Base URL behaves differently between OpenAI and Anthropic.  
I've already tried to explain it here: https://github.com/coder/aibridge/blob/b97c9d4fa4ffea26b27b54c2723375063bc67511/provider/provider.go#L14-L45  
and here is comment showing differences from PR: https://github.com/coder/aibridge/pull/174#discussion_r2782320152

provider/copilot.go
```diff
@@ -52,6 +52,9 @@ type Copilot struct {
 var _ Provider = &Copilot{}
 
 func NewCopilot(cfg config.Copilot) *Copilot {
+	if cfg.Name == "" {
```
pawbana, line 55: I know this is very simple but tests are missing

config/config.go
ssncferreira: Updated the comment to reference the name "anthropic", "openai", "copilot". 
On moving the defaulting logic to config: I'd prefer to keep it in the provider constructors for consistency, that's the same pattern used for the other methods like BaseURL, Key, APIDumpDir, and CircuitBreaker.

ssncferreira:
> > `RoutePrefix()` is derived from `Name()`
> 
> I'm not sure how I fell about this. If `Name` is configurable I think route prefix should also be configurable but maybe default could be based on `Name`to simplify configuration / UX.
> 
> If `RoutePrefix` would be configurable it needs good explanation how it works. Due to how Base URL behaves differently between OpenAI and Anthropic. I've already tried to explain it here:
> 
> https://github.com/coder/aibridge/blob/b97c9d4fa4ffea26b27b54c2723375063bc67511/provider/provider.go#L14-L45
> 
> 
> and here is comment showing differences from PR: [#174 (comment)](https://github.com/coder/aibridge/pull/174#discussion_r2782320152)

`RoutePrefix` is derived from `Name`, same as before, the difference is that `Name` is now configurable. This is intentional: the route prefix is what distinguishes provider instances of the same type:
* `https://coder.host.com/api/v2/aibridge/copilot/{route}`
* `https://coder.host.com/api/v2/aibridge/copilot-business/{route}`

I don't think `RoutePrefix` needs its own config field right now, if someone sets `Name: "copilot-business"`, the route prefix `/copilot-business` is the obvious correct default (we can add validation to the name to avoid additional `/` if needed). We can always add an explicit `RoutePrefix` override later if a use case comes up where the name and route need to diverge.

Your comment about OpenAI vs Anthropic base URL differences still applies: that behavior is unchanged since each provider's `RoutePrefix()` implementation handles it (e.g. OpenAI appends /v1).

provider/copilot.go
ssncferreira: Done 👍

dannykopping (approved):

pawbana (approved):

provider/anthropic.go
```diff
@@ -68,10 +71,14 @@ func NewAnthropic(cfg config.Anthropic, bedrockCfg *config.AWSBedrock) *Anthropi
 	}
 }
 
-func (p *Anthropic) Name() string {
+func (p *Anthropic) Type() string {
 	return config.ProviderAnthropic
 }
 
+func (p *Anthropic) Name() string {
+	return p.cfg.Name
```
pawbana, line 79: Since `Name` defines route prefix does it need to be checked/sanitized?

ssncferreira: Yes, it should. I can add it in `validateProviders` on the upstack PR: https://github.com/coder/aibridge/pull/240

ssncferreira: Added here: https://github.com/coder/aibridge/pull/240/commits/7d0c1986578b12e2ff91ae78c08665876838edff

ssncferreira:
### Merge activity

* **Mar 31, 2:32 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/coder/aibridge/237).
* **Mar 31, 2:33 PM UTC**: [Graphite](https://app.graphite.com/github/pr/coder/aibridge/237) rebased this pull request as part of a merge.
* **Mar 31, 2:33 PM UTC**: @ssncferreira merged this pull request with [Graphite](https://app.graphite.com/github/pr/coder/aibridge/237).

PR #238, opened 2026-03-30
fix: pass provider name to interceptors instead of hardcoding it

ssncferreira:
## Problem

Interceptors hardcode the provider name for API dump and trace attributes. The `chatcompletions` and `responses` interceptors always used "openai", and the messages interceptor always used "anthropic". For Copilot, since it uses the OpenAI interceptor, this meant that requests were dumped to the `openai/` directory and traced as `provider=openai` instead of `copilot`.

## Changes

* Add `providerName` parameter to all interceptor constructors
* Replace hardcoded provider constants with the passed provider name in API dump and trace attributes
* Providers pass `p.Name()` when constructing interceptors
* Add Copilot API dump integration tests verifying dumps land in the correct `copilot/` directory

Related to: https://github.com/coder/aibridge/issues/152

_Disclaimer: initially produced by Claude Opus 4.6, heavily modified and reviewed by @ssncferreira ._

PR #240, opened 2026-03-30
feat: validate duplicate provider names in NewRequestBridge

ssncferreira:
## Description

Adds validation to `NewRequestBridge` to prevent registering multiple providers with the same name. Duplicate names would cause route prefix conflicts and ambiguous metrics, logs, and circuit breaker behavior.

## Changes

* Add `validateProviders()` that rejects duplicate provider names at startup
* Add table-driven tests covering valid configurations and duplicate name detection

Related to: https://github.com/coder/aibridge/issues/152

_Disclaimer: initially produced by Claude Opus 4.6, heavily modified and reviewed by @ssncferreira ._

### DIFF
--- a/bridge.go
+++ b/bridge.go
@@ -57,6 +57,19 @@ type RequestBridge struct {
 
 var _ http.Handler = &RequestBridge{}
 
+// validateProviders checks that no two providers share the same name.
+func validateProviders(providers []provider.Provider) error {
+	names := make(map[string]bool, len(providers))
+	for _, prov := range providers {
+		if names[prov.Name()] {
+			return fmt.Errorf("duplicate provider name: %q", prov.Name())
+		}
+		names[prov.Name()] = true
+	}
+	// TODO(ssncferreira): validate duplicate baseURLs as well
+	return nil
+}
+
 // NewRequestBridge creates a new *[RequestBridge] and registers the HTTP routes defined by the given providers.
 // Any routes which are requested but not registered will be reverse-proxied to the upstream service.
 //
@@ -67,6 +80,10 @@ var _ http.Handler = &RequestBridge{}
 // Circuit breaker configuration is obtained from each provider's CircuitBreakerConfig() method.
 // Providers returning nil will not have circuit breaker protection.
 func NewRequestBridge(ctx context.Context, providers []provider.Provider, rec recorder.Recorder, mcpProxy mcp.ServerProxier, logger slog.Logger, m *metrics.Metrics, tracer trace.Tracer) (*RequestBridge, error) {
+	if err := validateProviders(providers); err != nil {
+		return nil, err
+	}
+
 	mux := http.NewServeMux()
 
 	for _, prov := range providers {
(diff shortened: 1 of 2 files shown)


ssncferreira:
* **#240** <a href="https://app.graphite.com/github/pr/coder/aibridge/240?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/aibridge/240?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#238** <a href="https://app.graphite.com/github/pr/coder/aibridge/238?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#237** <a href="https://app.graphite.com/github/pr/coder/aibridge/237?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

bridge.go
```diff
@@ -57,6 +57,18 @@ type RequestBridge struct {
...
+// validateProviders checks that no two providers share the same name.
+func validateProviders(providers []provider.Provider) error {
+	names := make(map[string]bool, len(providers))
+	for _, prov := range providers {
+		if names[prov.Name()] {
+			return fmt.Errorf("duplicate provider name: %q", prov.Name())
+		}
+		names[prov.Name()] = true
+	}
+	return nil
```
ssncferreira, line 77: Initially, I added a check to confirm that there are no 2 providers with the same base URL. However, this was failing on the tests, because `newBridgeTestServer` receives a single upstreamURL, which is then used for both anthropic and openai: https://github.com/coder/aibridge/blob/main/internal/integrationtest/setupbridge.go#L163-L164

We might need to update the tests, but that would require a bigger refactor.

pawbana: Maybe I'm missing something but I don't think we should care about base url being the same?

The only issue is prefix provider being mounted on being the same which is checked.

I could imagine this being useful, eg. having the same provider with the same base URL (default value) with different keys behind them and different permissions to them. Something like per org provider of the same type/upstream with different limits.

ssncferreira: That is a good point, I didn't think of it. Removed the TODO: https://github.com/coder/aibridge/pull/240/commits/225fdde21903935b8e234e7e61e82ccf2ff8cf64

dannykopping (approved):
The timing for this is funny because I have to look into https://linear.app/codercom/issue/AIGOV-157/spike-to-understand-if-there-is-a-simple-way-to-handle-multi-api-key which may _require_ multiple providers of the same name, but I'll have to find a way around this.

pawbana (approved):

ssncferreira:
### Merge activity

* **Mar 31, 2:32 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/coder/aibridge/240).
* **Mar 31, 2:36 PM UTC**: [Graphite](https://app.graphite.com/github/pr/coder/aibridge/240) rebased this pull request as part of a merge.
* **Mar 31, 2:37 PM UTC**: @ssncferreira merged this pull request with [Graphite](https://app.graphite.com/github/pr/coder/aibridge/240).

PR #241, opened 2026-03-31
fix: handle Claude Code session detection for new header / metadata format

dannykopping:
Claude Code changed this metadata format (from which we extract the session ID):

before:

```json
...
  "metadata": {
    "user_id": "user_..._account__session_57a121ec-f78c-48c3-a745-e3036a3ecc33"
  },
...
```

after:

```json
...
  "metadata": {
    "user_id": "{\"device_id\":\"...\",\"account_uuid\":\"\",\"session_id\":\"54c1eb09-bc4c-4d2f-98eb-6d2ab2d5e2fe\"}"
  },
...
```

To make matters even more complicated, from v2.1.86 it introduced as `X-Claude-Code-Session-Id` header.

This PR supports all 3 approaches.

### DIFF
--- a/session.go
+++ b/session.go
@@ -11,7 +11,7 @@ import (
 	"github.com/tidwall/gjson"
 )
 
-var claudeCodePattern = regexp.MustCompile(`_session_(.+)$`) // Save compilation on each call.
+var claudeCodePattern = regexp.MustCompile(`_session_(.+)$`) // Legacy format: save compilation on each call.
 
 // guessSessionID attempts to retrieve a session ID which may have been sent by
 // the client. We only attempt to retrieve sessions using methods recognized for
@@ -20,12 +20,17 @@ func guessSessionID(client Client, r *http.Request) *string {
 	switch client {
 	case ClientClaudeCode:
 		/* Claude Code adds the session ID into the `metadata.user_id` field in the JSON body.
+		Legacy format:
 		{
-			...
 			"metadata": {
 				"user_id": "user_{sha256}_account_{account_id}_session_{uuid_v4}"
-			},
-			...
+			}
+		}
+		New format (the value is a JSON-encoded string):
+		{
+			"metadata": {
+				"user_id": "{\"device_id\":\"...\",\"account_uuid\":\"...\",\"session_id\":\"uuid_v4\"}"
+			}
 		} */
 		payload, err := io.ReadAll(r.Body)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


session.go
```diff
@@ -11,25 +11,24 @@ import (
...
-			...
-		} */
+		// Prefer the dedicated header (added in Claude Code v2.1.86+).
+		if sid := cleanRef(r.Header.Get("X-Claude-Code-Session-Id")); sid != nil {
+			return sid
+		}
+
+		// Fall back to extracting from the metadata.user_id field in the JSON body.
+		// Legacy format: "user_{sha256}_account_{id}_session_{uuid}"
+		// Newer format:  JSON-encoded object with a "session_id" field.
```
ssncferreira: nit: should we flip this comment? IIUC, first we try the newer format and then the legacy one

session_test.go
```diff
@@ -21,12 +21,49 @@ func TestGuessSessionID(t *testing.T) {
...
+		},
+		{
+			name:      "claude_code_header_only",
+			client:    ClientClaudeCode,
+			headers:   map[string]string{"X-Claude-Code-Session-Id": "aabb-ccdd"},
+			body:      `{"model":"claude-3"}`,
+			sessionID: utils.PtrTo("aabb-ccdd"),
+		},
+		{
+			name:      "claude_code_empty_header_falls_back_to_body",
```
ssncferreira, line 39: nit: another one `claude_code_whitespace_header_falls_back_to_body`

ssncferreira (approved):
LGTM 🏆

session.go
pawbana: nit: could check`userID.Type != gjson.String`

session_test.go
```diff
@@ -21,12 +21,49 @@ func TestGuessSessionID(t *testing.T) {
 		sessionID *string
 	}{
 		// Claude Code.
+		{
+			name:      "claude_code_header_takes_precedence",
+			client:    ClientClaudeCode,
+			headers:   map[string]string{"X-Claude-Code-Session-Id": "header-session-id"},
+			body:      `{"metadata":{"user_id":"user_abc123_account_456_session_body-session-id"}}`,
+			sessionID: utils.PtrTo("header-session-id"),
```
pawbana, line 29: nit: `sessionID` -> `expectSessionID` would make it clear what is input and what is output / needs to be checked.

pawbana (approved):

session_test.go
dannykopping: I'd rather keep this change-set small; I think it's fine for now.

PR #242, opened 2026-03-31
chore: add support for Coder Agents sessions

dannykopping:
_Disclaimer:_ _implemented_ _using_ _Claude_ _Opus_ _4\.6, reviewed by me._

Follow-up from https://github.com/coder/coder/pull/23578

We still need to fake the initiator using the `X-Coder-Owner-Id`; that's implemented upstack.
Validated manually using Coder Agents with an OpenAI-Compatible provider pointed at AI Bridge in my workspace.

PR #244, opened 2026-03-31
feat: prefer X-Coder-Owner-Id header when present

dannykopping:
_Disclaimer:_ _implemented_ _using_ _Claude_ _Opus_ _4\.6, reviewed by me._

When AI Bridge is used in conjunction with Coder Agents, it passes along a header (`X-Coder-Owner-Id`) which specifies the user which initiated the interaction with Coder Agents (and transitively with AI Bridge).

We should prefer this ID over the initiator passed through in the auth credential because Coder Agents can currently only provide a single token for all users when using an OpenAI-compatible provider. If a service account credential is used, we don't want the service account to show up as the initiator but rather the actual end-user.

This header was introduced in https://github.com/coder/coder/pull/23578.

Manually tested and it works as expected.

### DIFF
--- a/bridge.go
+++ b/bridge.go
@@ -13,6 +13,7 @@ import (
 
 	"cdr.dev/slog/v3"
 	"github.com/coder/aibridge/circuitbreaker"
+	"github.com/google/uuid"
 	aibcontext "github.com/coder/aibridge/context"
 	"github.com/coder/aibridge/mcp"
 	"github.com/coder/aibridge/metrics"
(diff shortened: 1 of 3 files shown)


dannykopping:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/coder/aibridge/244?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#244** <a href="https://app.graphite.com/github/pr/coder/aibridge/244?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/aibridge/244?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#242** <a href="https://app.graphite.com/github/pr/coder/aibridge/242?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

bridge.go
```diff
@@ -172,6 +172,27 @@ func newInterceptionProcessor(p provider.Provider, cbs *circuitbreaker.ProviderC
 			}()
 		}
 
+		// For Coder Agents, the X-Coder-Owner-Id header identifies the actual
+		// user who initiated the chat. Override the actor so usage is attributed
+		// to the correct user rather than the service-level identity.
+		if client == ClientCoderAgents {
+			if ownerID := r.Header.Get("X-Coder-Owner-Id"); ownerID != "" {
```
dannykopping, line 204: We have to consider that it's _possible_ that someone malicious could spoof the User-Agent and the `X-Coder-Owner-Id` header here. We're validating that it's at least a valid UUIDv4 value to prevent XSS, so the best they could do is make their interceptions masquerade as someone else's, but we still log this change.

PR #267, opened 2026-04-15
chore: create proxy instance once per provider

### LINKED ISSUE #187: Create proxy per pass though route once
Currently new proxy for pass though routes is created for each request.
It should be created only once for each route.

pawbana:
Changes `newPassthroughRouter` so it created proxy per provider not on each request.
Replaced `Director` in `httputil.ReverseProxy` with `Rewrite` as Director is being [deprecated](https://cs.opensource.google/go/go/+/refs/tags/go1.26.2:src/net/http/httputil/reverseproxy.go;l=185) in next go version.

Fixes: https://github.com/coder/aibridge/issues/187

### DIFF
--- a/passthrough.go
+++ b/passthrough.go
@@ -1,7 +1,6 @@
 package aibridge
 
 import (
-	"net"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
@@ -21,7 +20,39 @@ import (
 
 // newPassthroughRouter returns a simple reverse-proxy implementation which will be used when a route is not handled specifically
 // by a [intercept.Provider].
+// A single reverse proxy is created per provider and reused across all requests.
 func newPassthroughRouter(prov provider.Provider, logger slog.Logger, m *metrics.Metrics, tracer trace.Tracer) http.HandlerFunc {
+	provBaseURL, err := url.Parse(prov.BaseURL())
+	if err != nil {
+		return newInvalidBaseURLHandler(prov, logger, m, tracer, err)
+	}
+	if _, err := url.JoinPath(provBaseURL.Path, "/"); err != nil {
+		return newInvalidBaseURLHandler(prov, logger, m, tracer, err)
+	}
+
+	// Transport tuned for streaming (no response header timeout).
+	t := &http.Transport{
+		Proxy:                 http.ProxyFromEnvironment,
+		ForceAttemptHTTP2:     true,
+		MaxIdleConns:          100,
+		IdleConnTimeout:       90 * time.Second,
+		TLSHandshakeTimeout:   10 * time.Second,
+		ExpectContinueTimeout: 1 * time.Second,
+	}
+
+	// Build a reverse proxy to the upstream, reused across all requests for this provider.
+	// All request modifications happen in Rewrite.
+	proxy := &httputil.ReverseProxy{
+		Rewrite: func(pr *httputil.ProxyRequest) {
+			rewritePassthroughRequest(pr, provBaseURL, prov)
+		},
+		Transport: apidump.NewPassthroughMiddleware(t, prov.APIDumpDir(), prov.Name(), logger, quartz.NewReal()),
+		ErrorHandler: func(rw http.ResponseWriter, req *http.Request, e error) {
+			logger.Warn(req.Context(), "reverse proxy error", slog.Error(e), slog.F("path", req.URL.Path))
+			http.Error(rw, "upstream proxy error", http.StatusBadGateway)
+		},
+	}
+
 	return func(w http.ResponseWriter, r *http.Request) {
 		if m != nil {
 			m.PassthroughCount.WithLabelValues(prov.Name(), r.URL.Path, r.Method).Add(1)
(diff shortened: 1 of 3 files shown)


passthrough.go
```diff
@@ -21,7 +20,39 @@ import (
 
 // newPassthroughRouter returns a simple reverse-proxy implementation which will be used when a route is not handled specifically
 // by a [intercept.Provider].
+// A single reverse proxy is created per provider and reused across all requests.
 func newPassthroughRouter(prov provider.Provider, logger slog.Logger, m *metrics.Metrics, tracer trace.Tracer) http.HandlerFunc {
+	provBaseURL, err := url.Parse(prov.BaseURL())
+	if err != nil {
+		return newInvalidBaseURLHandler(prov, logger, m, tracer, err)
+	}
+	if _, err := url.JoinPath(provBaseURL.Path, "/"); err != nil {
```
ssncferreira, line 30: Why is this needed? If `provBaseURL` is properly set, this will never fail, right? 🤔

passthrough.go
```diff
@@ -33,88 +64,48 @@ func newPassthroughRouter(prov provider.Provider, logger slog.Logger, m *metrics
...
-			},
-			ErrorHandler: func(rw http.ResponseWriter, req *http.Request, e error) {
-				logger.Warn(req.Context(), "reverse proxy error", slog.Error(e), slog.F("path", req.URL.Path))
-				http.Error(rw, "upstream proxy error", http.StatusBadGateway)
-			},
-		}
+	if prior, ok := pr.In.Header["X-Forwarded-For"]; ok {
+		pr.Out.Header["X-Forwarded-For"] = append([]string(nil), prior...)
+	}
+	pr.SetXForwarded()
```
ssncferreira, line 81: Slightly confused with this, IIUC SetXForwarded sets all `X-Forwarded` headers right? `For`, `Host` and `Proto`, so why do we need the first if?

nit: might be worth adding a small comment

passthrough.go
```diff
@@ -33,88 +64,48 @@ func newPassthroughRouter(prov provider.Provider, logger slog.Logger, m *metrics
...
+		pr.Out.Header.Set("User-Agent", "aibridge") // TODO: use build tag.
+	}
+
+	// Inject provider auth.
+	prov.InjectAuthHeader(&pr.Out.Header)
+}
+
+// newInvalidBaseURLHandler returns a handler that always returns 502 because
+// the provider's base URL is invalid.
+func newInvalidBaseURLHandler(prov provider.Provider, logger slog.Logger, m *metrics.Metrics, tracer trace.Tracer, baseURLErr error) http.HandlerFunc {
```
ssncferreira, line 97: Nit: this duplicates the metrics/tracing from the happy-path handler (https://github.com/coder/aibridge/pull/267/changes#diff-b2b968f6cc08ffeda242a5870824976aa457930e6c2ef8b0ea80e747540dcd56R56). 
Maybe we could extract the common setup into a wrapper.

passthrough_test.go
```diff
@@ -85,3 +130,161 @@ func TestPassthroughRoutes(t *testing.T) {
...
+				"X-Forwarded-For":   {"2.2.2.2, 3.3.3.3, 1.1.1.1"},
+				"User-Agent":        {"aibridge"},
+			},
+		},
+		{
+			name:          "omits_forwarded_for_when_remote_addr_is_not_parseable",
+			reqPath:       "http://client-host/chat",
+			reqRemoteAddr: "not-a-socket-address",
+			reqHeaders: http.Header{
+				"X-Forwarded-For": {"1.1.1.1"},
```
ssncferreira, line 233: Why isn't this header in the `expectHeader` set?

passthrough_test.go
```diff
@@ -85,3 +130,161 @@ func TestPassthroughRoutes(t *testing.T) {
...
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			r := httptest.NewRequest(http.MethodGet, tc.reqPath, nil)
+			maps.Copy(r.Header, tc.reqHeaders)
+			r.RemoteAddr = tc.reqRemoteAddr
+			if tc.reqTLS {
```
ssncferreira, line 252: nit: I don't think we have a test where this is true 🤔

ssncferreira (commented):
Overall LGTM, just some questions to make sure I understand this change 🙂

passthrough.go
pawbana: This one is interesting, `url.Parse` doesn't verify that path part of URL is properly encoded, that is what `url.JoinPath` is for.
It may be a bit pedantic but was already tested in `passthrough_route_break_parse_base_url` and I wanted to keep checks the same.

passthrough.go
pawbana: From [documentation](https://pkg.go.dev/net/http/httputil#ProxyRequest.SetXForwarded)

> To append to the inbound request's X-Forwarded-For header (the default behavior of [ReverseProxy](https://pkg.go.dev/net/http/httputil#ReverseProxy) when using a Director function), copy the header from the inbound request before calling SetXForwarded

I'll add a comment

passthrough_test.go
pawbana: As test case name suggest: `omits_forwarded_for_when_remote_addr_is_not_parseable`

`reqRemoteAddr` is set to not proper value. This is weird edge case that probably will never happen but I think it is better to point it out.

passthrough_test.go
pawbana: oh good catch, must have removed it by mistake. Will add back.

passthrough_test.go
ssncferreira: Right, but `1.1.1.1` is a valid address; it should still be kept in the header, no?

pawbana: I'm not exactly sure why library decides to drop the header but maybe due to a security concerns?
If RemoteAddr / last hop is unknown it is better to drop whole header then forward it with missing hop.
Maybe I should remove this test as it tests library rather then our code.

ssncferreira: Yeah, I think this is a very edge case and not sure it can even occur in production. I would suggest either removing the test, or if we decide to keep it add a comment explaining this for future reference

ssncferreira (approved):
LGTM 💪 thank you for addressing the comments!

PR #269, opened 2026-04-17
feat: log warning when SSE flush to client is slow

### LINKED ISSUE #66: Slow client detection
Add some way to identify when sending data from AI Bridge to client is slow. https://github.com/coder/aibridge/blob/9e9491bee298a99a6b53c38a62e1269b5a5a042b/streaming.go#L56

Possible solutions:
- trace every flush
  - very detailed but too many traces
- trace whole `start` method
  - not much detail, long flush in the middle will not be visible, cheap
  - maybe add events on each flush start/end?
- filter spans by duration
  - would require custom exporter?
- log when flush takes a long time
  - easy but not part of tracing

mtojek:
Fixes: #66 

Adds slow client detection to the EventStream SSE sender. After each flush, the elapsed time is measured using an injected quartz.Clock. If it exceeds 500ms, a Warn log is emitted with the flush duration, making slow clients immediately visible in logs without requiring distributed tracing infrastructure.

quartz.Clock is injected into EventStream so tests can simulate slow flushes by advancing a mock clock — no real sleeping required.

### DIFF
--- a/intercept/eventstream/eventstream.go
+++ b/intercept/eventstream/eventstream.go
@@ -150,7 +150,7 @@ func (s *EventStream) Start(w http.ResponseWriter, r *http.Request) {
 				slog.F("flush_duration", d),
 				slog.F("client_ip", clientIP),
 				slog.F("user_agent", r.Header.Get("User-Agent")),
-				slog.F("response_size_bytes", len(ev)),
+				slog.F("payload_size", len(ev)),
 			)
 		}
 
--- a/intercept/eventstream/eventstream.go
+++ b/intercept/eventstream/eventstream.go
@@ -21,8 +21,10 @@ import (
 var ErrEventStreamClosed = xerrors.New("event stream closed")
 
 const (
-	pingInterval       = time.Second * 10
-	slowFlushThreshold = time.Millisecond * 500
+	pingInterval = time.Second * 10
+	// SlowFlushThreshold is the duration after which a flush to the client is
+	// considered slow and a warning is logged.
+	SlowFlushThreshold = time.Millisecond * 500
 )
 
 type event []byte
@@ -142,7 +144,7 @@ func (s *EventStream) Start(w http.ResponseWriter, r *http.Request) {
 			s.logger.Warn(ctx, "failed to flush event stream", slog.Error(err))
 			return
 		}
-		if d := s.clk.Since(flushStart); d > slowFlushThreshold {
+		if d := s.clk.Since(flushStart); d > SlowFlushThreshold {
 			s.logger.Warn(ctx, "slow client detected", slog.F("flush_duration", d))
 		}
 
--- a/intercept/eventstream/eventstream.go
+++ b/intercept/eventstream/eventstream.go
@@ -145,7 +145,13 @@ func (s *EventStream) Start(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 		if d := s.clk.Since(flushStart); d > SlowFlushThreshold {
-			s.logger.Warn(ctx, "slow client detected", slog.F("flush_duration", d))
+			clientIP, _, _ := net.SplitHostPort(r.RemoteAddr)
+			s.logger.Warn(ctx, "slow client detected",
+				slog.F("flush_duration", d),
+				slog.F("client_ip", clientIP),
+				slog.F("user_agent", r.Header.Get("User-Agent")),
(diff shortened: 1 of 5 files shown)


intercept/eventstream/eventstream.go
```diff
@@ -131,10 +137,14 @@ func (s *EventStream) Start(w http.ResponseWriter, r *http.Request) {
 			}
 			return
 		}
+		flushStart := s.clk.Now()
 		if err := flush(w); err != nil {
 			s.logger.Warn(ctx, "failed to flush event stream", slog.Error(err))
 			return
 		}
+		if d := s.clk.Since(flushStart); d > slowFlushThreshold {
+			s.logger.Warn(ctx, "slow client detected", slog.F("flush_duration", d))
```
dannykopping: This won't really be enough information for an operator to respond to. What is the operator reasonably expected to do in this case (i.e. how can we make this actionable)?

intercept/eventstream/eventstream_test.go
```diff
@@ -0,0 +1,133 @@
...
+// Hijack satisfies the FullResponseWriter lint rule.
+func (*clockAdvancingFlusher) Hijack() (net.Conn, *bufio.ReadWriter, error) {
+	return nil, nil, nil
+}
+
+func TestEventStream_LogsWarning_WhenFlushIsSlow(t *testing.T) {
+	t.Parallel()
+
+	sink := &captureSink{}
+	logger := slogtest.Make(t, nil).AppendSinks(sink).Leveled(slog.LevelWarn)
```
dannykopping: You can simplify this PR a bit by using `sloghuman.Sink` with a `strings.Builder` (see `agent/agentcontainers/api_test.go`)

intercept/eventstream/eventstream_test.go
```diff
@@ -0,0 +1,133 @@
...
+	ctx := context.Background()
+	clk := quartz.NewMock(t)
+
+	stream := eventstream.NewEventStream(ctx, logger, nil, clk)
+
+	w := &clockAdvancingFlusher{
+		ResponseRecorder: httptest.NewRecorder(),
+		clk:              clk,
+		advance:          600 * time.Millisecond, // exceeds slowFlushThreshold (500ms)
```
dannykopping: Rather reuse the threshold const here please.

dannykopping (commented):
Mostly looks good, but this log isn't actionable right now.

intercept/eventstream/eventstream.go
mtojek: Good point. 

We can make this more actionable by including more context, such as user-agent (if available), and response size. This would make it easier to distinguish between an isolated slow client and a broader systemic issue.

Is that what you meant?

intercept/eventstream/eventstream_test.go
mtojek: TIL 👍

.claude/settings.local.json
```diff
@@ -0,0 +1,22 @@
+{
```
dannykopping: Please remove this.

intercept/eventstream/eventstream.go
dannykopping: That sounds like a good start to me 👍 
If we can include the client's IP address that'd also help close the loop, potentially.

mtojek: Done, thanks for challenge

intercept/eventstream/eventstream.go
```diff
@@ -131,10 +139,20 @@ func (s *EventStream) Start(w http.ResponseWriter, r *http.Request) {
...
 			s.logger.Warn(ctx, "failed to flush event stream", slog.Error(err))
 			return
 		}
+		if d := s.clk.Since(flushStart); d > SlowFlushThreshold {
+			clientIP, _, _ := net.SplitHostPort(r.RemoteAddr)
+			s.logger.Warn(ctx, "slow client detected",
+				slog.F("flush_duration", d),
+				slog.F("client_ip", clientIP),
+				slog.F("user_agent", r.Header.Get("User-Agent")),
+				slog.F("response_size_bytes", len(ev)),
```
dannykopping: Nit: this isn't the full response, but rather a partial SSE event.

```suggestion
				slog.F("payload_size", len(ev)),
```

dannykopping (approved):
LGTM, just one nit
I don't need to review again
Thanks a lot @mtojek!

PR #273, opened 2026-04-17
fix(intercept/responses): record all tool call types in responses interceptor

### LINKED ISSUE #121: Consider recording of other tool types in responses interceptor
Currently Responses API interceptor only records function calls and custom calls.

Consider if more types of tool calls should be recorded:
* [computer_call](https://platform.openai.com/docs/guides/tools-computer-use) | [item](https://platform.openai.com/docs/api-reference/responses/object#responses-object-output-computer_tool_call-type)
* [local_shell_call](https://platform.openai.com/docs/guides/tools-local-shell) | [item](https://platform.openai.com/docs/api-reference/responses/object#responses-object-output-local_shell_call-type)
* [shell_call](https://platform.openai.com/docs/guides/tools-shell) | [item](https://platform.openai.com/docs/api-reference/responses/object#responses-object-output-shell_tool_call-type)
* [apply_patch_call](https://platform.openai.com/docs/guides/tools-apply-patch) | [item](https://platform.openai.com/docs/api-reference/responses/object#responses-object-output-apply_patch_tool_call-type)

Not handled by client but maybe would be useful for observability:
* [code_interpreter_call](https://platform.openai.com/docs/guides/tools-code-interpreter) | [item](https://platform.openai.com/docs/api-reference/responses/object#responses-object-output-code_interpreter_tool_call-type)
* [mcp_call](https://developers.openai.com/cookbook/examples/mcp/mcp_tool_guide) | [item](https://platform.openai.com/docs/api-reference/responses/object#responses-object-output-mcp_tool_call-type)

Probably not interesting:
* [file_search_call](https://platform.openai.com/docs/guides/tools-file-search) | [item](https://platform.openai.com/docs/api-reference/responses/object#responses-object-output-file_search_tool_call-type)
* [web_search_call ?](https://platform.openai.com/docs/guides/tools-web-search) | [item](https://platform.openai.com/docs/api-reference/responses/object#responses-object-output-web_search_tool_call-type)
* [image_generation_call ?](https://platform.openai.com/docs/guides/tools-image-generation) | [item](https://platform.openai.com/docs/api-reference/responses/object#responses-object-output-image_generation_call-type)

dannykopping:
*Disclaimer: implemented by a Coder Agent using Claude Opus 4.6*

## Summary

Previously, the responses interceptor only recorded `function_call` and `custom_tool_call` types. This left tool calls like `web_search_call`, `computer_call`, `shell_call`, etc. unrecorded — meaning interceptions that did significant work via built-in tools showed no tool usage at all.

Now all tool call types are recorded for maximum auditor visibility:

| Tool type | Previously recorded | Now recorded |
|---|---|---|
| `function_call` | ✅ | ✅ |
| `custom_tool_call` | ✅ | ✅ |
| `web_search_call` | ❌ | ✅ |
| `computer_call` | ❌ | ✅ |
| `local_shell_call` | ❌ | ✅ |
| `shell_call` | ❌ | ✅ |
| `apply_patch_call` | ❌ | ✅ |
| `code_interpreter_call` | ❌ | ✅ |
| `mcp_call` | ❌ | ✅ |
| `file_search_call` | ❌ | ✅ |
| `image_generation_call` | ❌ | ✅ |

## Approach

For the new tool types:
- **Tool name** falls back to the `type` field when no explicit `name` is set (most built-in tools don't have a name, unlike function calls).
- **Call ID** falls back to `id` when `call_id` is empty (`web_search_call`, `file_search_call`, `code_interpreter_call`, `image_generation_call` use `id` instead of `call_id`).
- Non-tool output types (`message`, `reasoning`, `*_output`, etc.) are still skipped.

## Changes

- `intercept/responses/base.go` — expanded `recordNonInjectedToolUsage` switch to handle all tool call types.
- `intercept/responses/base_test.go` — added `web_search_call_with_no_name` and `all_additional_tool_types` test cases.
- `fixtures/openai/responses/blocking/web_search.txtar` — new fixture for a web search response.
- `fixtures/fixtures.go` — embedded the new fixture.

Closes https://github.com/coder/aibridge/issues/121
Linear: AIGOV-96

### DIFF
--- a/intercept/responses/base.go
+++ b/intercept/responses/base.go
@@ -34,10 +34,6 @@ import (
 
 const (
 	requestTimeout = time.Second * 600
-
-	// computerCall is defined locally because the OpenAI Go SDK does
-	// not export a typed constant for the "computer_call" output item.
-	computerCall = "computer_call"
 )
 
 type responsesInterceptionBase struct {
@@ -227,7 +223,7 @@ func (i *responsesInterceptionBase) recordNonInjectedToolUsage(ctx context.Conte
 
 		// Agentic tools: the client sends a corresponding *_output
 		// item correlated by call_id.
-		case computerCall,
+		case "computer_call",
 			string(constant.ValueOf[constant.LocalShellCall]()),
 			string(constant.ValueOf[constant.ShellCall]()),
 			string(constant.ValueOf[constant.ApplyPatchCall]()):
--- a/intercept/responses/base.go
+++ b/intercept/responses/base.go
@@ -204,14 +204,45 @@ func (i *responsesInterceptionBase) recordNonInjectedToolUsage(ctx context.Conte
 	}
 
 	for _, item := range response.Output {
-		var args recorder.ToolArgs
+		var (
+			args     recorder.ToolArgs
+			toolName string
+			callID   string
+		)
 
-		// recording other function types to be considered: https://github.com/coder/aibridge/issues/121
+		//nolint:goconst // tool type string values are clearer inline.
 		switch item.Type {
 		case string(constant.ValueOf[constant.FunctionCall]()):
 			args = i.parseFunctionCallJSONArgs(ctx, item.Arguments)
+			toolName = item.Name
+			callID = item.CallID
+
 		case string(constant.ValueOf[constant.CustomToolCall]()):
 			args = item.Input
+			toolName = item.Name
+			callID = item.CallID
+
+		case "web_search_call",
+			"computer_call",
+			"file_search_call",
+			"code_interpreter_call",
+			"image_generation_call",
+			"mcp_call",
+			"local_shell_call",
+			"shell_call",
+			"apply_patch_call":
+			// For these tool types, use the type as the tool name
+			// when there is no explicit name, and prefer call_id
+			// but fall back to the item id.
+			toolName = item.Name
+			if toolName == "" {
+				toolName = item.Type
+			}
+			callID = item.CallID
+			if callID == "" {
+				callID = item.ID
+			}
+
 		default:
 			continue
 		}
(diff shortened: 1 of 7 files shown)


intercept/responses/base.go
```diff
@@ -204,27 +204,63 @@ func (i *responsesInterceptionBase) recordNonInjectedToolUsage(ctx context.Conte
...
+		case string(constant.ValueOf[constant.WebSearchCall]()),
+			string(constant.ValueOf[constant.FileSearchCall]()),
+			string(constant.ValueOf[constant.CodeInterpreterCall]()),
+			string(constant.ValueOf[constant.ImageGenerationCall]()),
+			string(constant.ValueOf[constant.McpCall]()):
+			toolName = item.Name
+			if toolName == "" {
+				toolName = item.Type
+			}
+			callID = item.ID
```
pawbana: I don't like `callID` being assigned here to `item.ID`. Those are different types.
In theory it would be possible to decipher value type (item.callID / item.ID) based on function call type but I think in this form it would cause too much confusion.
Separate field would be clearer, eg `ToolCallItemID`.

dannykopping: All we're doing is storing these values. At the end of the day they're used to track particular instances of tool calls. Whether it's a call ID or a regular ID is kinda irrelevant; I think expanding the API (and therefore the DB) to accommodate this extra detail is unnecessary and an implementation detail of OpenAI specifically.

pawbana: Yes, for just storing some ID for identification purpose it is fine to do it this way, but If there is some other data source that correctly records fields it becomes hard / unclear why sometimes tool call ID doesn't match.
Maybe I'm too paranoid but I think we should be careful about data consistency in recordings.

dannykopping: Turns out tool calls actually have _two_ IDs: `id` and `call_id` 😆 
Right now we're throwing `id` on the floor
So maybe it is actually just better to capture both separately and avoid the correctness issues you're raising

I'll expand the scope of the ticket and get to this next week
Thanks 👍

dannykopping: Agreed — addressed in the latest commit (`34c9a19`). `ToolUsageRecord` now has a separate `ItemID` field (the output item's unique `id`, always present) alongside `ToolCallID` (the `call_id` used by agentic tools for correlation, empty for hosted tools).

Hosted tools now record `callID = ""` instead of conflating `item.ID` into `callID`.

The coder/coder side to plumb this through proto → DB is in a companion draft PR: https://github.com/coder/coder/pull/24562

_Disclaimer: implemented by a Coder Agent using Claude Opus 4.6_

PR #265, opened 2026-04-13
feat: support IAM role-based authentication for AWS Bedrock

ssncferreira:
## Description

Make AWS Bedrock static credentials optional. When `AccessKey` and `AccessKeySecret` are not set, AI Bridge falls back to the AWS SDK default credential chain, which supports IAM Roles (instance profiles, IRSA, ECS task roles), SSO, shared credentials/config files, and environment variables.

This unblocks organizations that don't use IAM Users and rely on IAM Roles for AWS resource access.

## Changes

- Remove the hard requirement for `AccessKey` and `AccessKeySecret`
- Fall back to the SDK default credential chain when static credentials are not provided
- Add `SessionToken` to the config for temporary credential support via static config
- Validate that only one of `AccessKey`/`AccessKeySecret` isn't set (partial config is a misconfiguration)
- Fail fast on startup if no credential source can be resolved

Closes: https://github.com/coder/aibridge/issues/144
Closes: https://linear.app/codercom/issue/AIGOV-67

PR #275, opened 2026-04-17
fix: remove SessionToken from Bedrock config

ssncferreira:
## Description

Removes `SessionToken` from the `AWSBedrock` config. The field was introduced in #265 but shouldn't have been exposed. AWS temporary credentials are short-lived and aibridge reads config once at startup with no refresh mechanism. A user configuring a session token via aibridge would have a deployment that silently breaks when credentials expire.

Users with temporary credentials should rely on the AWS SDK default credential chain instead: credential sources cannot be mixed, so all three values (access key, secret, session token) must come from outside aibridge (env vars, shared config files, IAM roles, etc.). The SDK handles credential refresh automatically.

Related to: #265 
Related to: https://github.com/coder/coder/pull/24346

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -30,7 +30,6 @@ type Anthropic struct {
 type AWSBedrock struct {
 	Region                     string
 	AccessKey, AccessKeySecret string
-	SessionToken               string
 	Model, SmallFastModel      string
 	// If set, requests will be sent to this URL instead of the default AWS Bedrock endpoint
 	// (https://bedrock-runtime.{region}.amazonaws.com).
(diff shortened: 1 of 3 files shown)


ssncferreira:
* **#275** <a href="https://app.graphite.com/github/pr/coder/aibridge/275?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/aibridge/275?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

dannykopping (approved):

PR #277, opened 2026-04-21
fix: track Charm Crush client and session ID

dannykopping:
*Disclaimer: implemented by a Coder Agent using Claude Opus 4.6*

## Summary

Fixes client detection and session ID tracking for the [Charm Crush](https://github.com/charmbracelet/crush) AI coding client.
https://github.com/charmbracelet/crush/pull/2403/changes#diff-f48445f706f26bd59d7bae605c4445d3c178237d2489223bf9c1a813417d9713 changed it from  `Charm Crush/...` to `Charm-Crush/...`

### Bug fix: User-Agent matching

The actual Crush user-agent is `Charm-Crush/{version} (https://charm.land/crush)` (hyphenated), but `GuessClient` only checked for `charm crush/` (space-separated). After lowercasing, `Charm-Crush/0.2.0` becomes `charm-crush/0.2.0`, which did not match the `charm crush/` prefix.

Now matches both formats for backwards compatibility.

### Session ID tracking

Adds an explicit `ClientCrush` case to `GuessSessionID`. Crush does not currently send a session ID header to upstream AI providers, so this returns `nil` (consistent with how `ClientZed`, `ClientRoo`, and `ClientCursor` are handled).

### Tests

- Added `charm_crush_hyphen` test case for `GuessClient` using the real user-agent format.
- Added `crush_returns_empty` test case for `GuessSessionID`.

### DIFF
--- a/client.go
+++ b/client.go
@@ -51,7 +51,7 @@ func GuessClient(r *http.Request) Client {
 		return ClientRoo
 	case strings.HasPrefix(userAgent, "coder-agents/"):
 		return ClientCoderAgents
-	case strings.HasPrefix(userAgent, "charm crush/"):
+	case strings.HasPrefix(userAgent, "charm crush/") || strings.HasPrefix(userAgent, "charm-crush/"):
 		return ClientCrush
 	case r.Header.Get("x-cursor-client-version") != "":
 		return ClientCursor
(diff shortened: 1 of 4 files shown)


dannykopping (commented):
Self-review ✅ 
I've validated this locally

